PR #5, opened 2020-07-22
Commit after changes done as per review

mmudigon:


tests/unit/modules/dcnm/fixtures/dcnm_intf_vpc_configs.json
```diff
@@ -354,10 +351,8 @@
 				"peer2_pcid": 2
 			},
 			"switch": [
-				[
 					"10.122.84.109",
-			"10.122.84.108"
-				]
+			    "10.122.84.108"
```
mikewiebe, line 355: Here and elsewhere update to use a private ip address.

tests/unit/modules/dcnm/fixtures/dcnm_intf_subint_configs.json
```diff
@@ -1,4 +1,15 @@
 {
+    "mock_fab_inv_data": {
+      "10.122.84.108": {
```
mikewiebe, line 3: Use private ip

tests/unit/modules/dcnm/fixtures/dcnm_intf_pc_configs.json
```diff
@@ -428,7 +449,7 @@
       "switch": [
         "10.122.84.108"
```
mikewiebe, line 450: Private ip

mmudigon:
Mike,

These IP addresses are used in UT scripts. Even these things are supposed to be private? Because these are not going to sent on the line, I thought it is ok to use these.  I changed all my IT scripts. Let me know. I can change these too.

Thanks

Regards
Mallik

tests/unit/modules/dcnm/fixtures/dcnm_intf_vpc_configs.json
mmudigon: Ok. Will modify the code to use private IP addresses

tests/unit/modules/dcnm/fixtures/dcnm_intf_pc_configs.json
mmudigon: Ok. Will modify the code to use private IP addresses

tests/unit/modules/dcnm/fixtures/dcnm_intf_subint_configs.json
mmudigon: Ok. Will modify the code to use private IP addresses

mikewiebe:
As discussed let's add an action plugin to warn users and terminate the task execution if they try and use the old playbook format for `profile` and vcp interface list.

plugins/modules/dcnm_interface.py
```diff
@@ -76,6 +76,9 @@
       profile_pc:
         description:
           - Object profile which must be included for port channel interface configurations.
+          - NOTE: Though the key shown here is 'profile_pc' the actual key to be used in playbook
```
mikewiebe, line 79: @mmudigon I had a chance to chat about this with Shangxin and we should instead update the description to include the type that this particular parameter is associated with.

The structure would look something like this:

```yaml
      profile:
        description:
          - Object profile used to include suboptions for configuring different interface types.
        suboptions:
          # ----------------- Common suboptions -----------------
          mode:
            description:
              - Interface mode
            choices: ['trunk', 'access', 'l3', 'monitor']
            type: str
            required: true
          # ----------------- Suboptions for type: pc -----------------
          members:
            description:
              - Member interfaces that are part of this port channel
              - Requires type parameter to be set to 'pc'
            type: list
            elements: str
            required: true
          # ----------------- Suboptions for type: vpc -----------------
          peer1_pcid:
            description:
              - Port channel identifier of first peer. If this object is not included, then the value defaults to the
                vPC identifier. This value cannot be changed once vPC is created
              - Requires type parameter to be set to 'vpc'
            type: int
            choices: [Min:1, Max:4096]
            default: Default value is the vPC port identifier

mmudigon:
Hi Mike,

If we can add comments in the DOC section, we can do as you said. Not sure about if we can add any comments there.

Thanks

Regards
Mallik

PR #6, opened 2020-07-24
DCNM VRF - fixed all issues/enhancemnets/ - UT - Modified all and add…

kharicha:
…ed new test cases/new file

Fixed all DCNM VRF - all enhancements for moving the vlan to parent vrf creation and vrf allocation by user and dcnm
Fixed existing issues in the code and doc
Modified all the UT test cases and added new test cases for the changes/fix/enhancements made
Added new file for query test cases.

### DIFF
--- a/plugins/modules/dcnm_vrf.py
+++ b/plugins/modules/dcnm_vrf.py
@@ -15,7 +15,6 @@
 # limitations under the License.
 
 import json
-import socket
 import time
 import copy
 from ansible_collections.cisco.dcnm.plugins.module_utils.network.dcnm.dcnm import \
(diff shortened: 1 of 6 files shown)


plugins/modules/dcnm_vrf.py
```diff
@@ -216,28 +215,29 @@
       vrf_id: 9008011
       vrf_template: Default_VRF_Universal
       vrf_extension_template: Default_VRF_Extension_Universal
+      vlan_id: 2000
```
mikewiebe, line 218: As discussed today let's add an action plugin to warn users if they specify the `vlan_id` under the `attach:` block and terminate the ansible task.

Issue #16, opened 2020-09-17
status of network should be included when query


dsx1123:
when query the network information,  in response, networkStatus should be included even it is not configurable:
**environment**:
```
ansible-playbook 2.9.13
  config file = /root/.ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /root/virtualenv/ansible/lib/python3.7/site-packages/ansible
  executable location = /root/virtualenv/ansible/bin/ansible-playbook
  python version = 3.7.3 (default, Apr  3 2019, 05:39:12) [GCC 8.3.0]

collection version: 1.0.0
```
**example of tasks:**
```yml
---
- name:  verify | check if network is deployed
  cisco.dcnm.dcnm_network:
    fabric: "{{ fabric }}"
    config:
      - net_name: network_app
        vrf_name: vrf_blue
    state: query
  register: net_query

- debug:
    msg: "{{ net_query }}"

```
 **result:**
```
PLAY [verify] ********************************************************************************************************************************************************************************************************************************

TASK [test : verify | check if network is deployed] ******************************************************************************************************************************************************************************************
Wednesday 16 September 2020  21:50:09 -0700 (0:00:00.117)       0:00:00.117 ***
ok: [172.25.74.53]

TASK [test : debug] **************************************************************************************************************************************************************************************************************************
Wednesday 16 September 2020  21:50:18 -0700 (0:00:08.534)       0:00:08.651 ***
ok: [172.25.74.53] => {
    "msg": {
        "ansible_facts": {
            "discovered_interpreter_python": "/usr/bin/python"
        },
        "changed": false,
        "diff": [],
        "failed": false,
        "response": [
            {
                "attach": [],
                "gw_ip_subnet": "192.168.2.1/24",
                "net_extension_template": "Default_Network_Extension_Universal",
                "net_id": 30001,
                "net_name": "network_app",
                "net_template": "Default_Network_Universal",
                "vlan_id": "",
                "vrf_name": "vrf_blue"
            }
        ]
    }
}

```

**expected result**
networkStatus of response is included in query result

mikewiebe:
https://github.com/CiscoDevNet/ansible-dcnm/pull/22

Issue #15, opened 2020-09-17
status of vrf should be included when query vrf


dsx1123:
when query the vrf information,  in respone, vrfStatus should be included even it is not configurable:
**environment**:
```
ansible-playbook 2.9.13
  config file = /root/.ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /root/virtualenv/ansible/lib/python3.7/site-packages/ansible
  executable location = /root/virtualenv/ansible/bin/ansible-playbook
  python version = 3.7.3 (default, Apr  3 2019, 05:39:12) [GCC 8.3.0]

collection version: 1.0.0
```
**example of playbook:**
```yml
---
- name: verify | check if all VRFs are deployed
  cisco.dcnm.dcnm_vrf:
    fabric: "{{ fabric }}"
    config:
      - vrf_name: vrf_blue 
    state: query
  register: result

- debug:
    msg: "{{ result }}"
```
 **result:**
```
PLAY [verify] ********************************************************************************************************************************************************************************************************************************

TASK [test : verify | check if all VRFs are deployed] ****************************************************************************************************************************************************************************************
Wednesday 16 September 2020  21:42:55 -0700 (0:00:00.123)       0:00:00.123 ***
ok: [172.25.74.53]

TASK [test : debug] **************************************************************************************************************************************************************************************************************************
Wednesday 16 September 2020  21:43:00 -0700 (0:00:04.930)       0:00:05.054 ***
ok: [172.25.74.53] => {
    "msg": {
        "ansible_facts": {
            "discovered_interpreter_python": "/usr/bin/python"
        },
        "changed": false,
        "diff": [],
        "failed": false,
        "response": [
            {
                "attach": [],
                "service_vrf_template": null,
                "source": null,
                "vrf_extension_template": "Default_VRF_Extension_Universal",
                "vrf_id": 50000,
                "vrf_name": "vrf_blue",
                "vrf_template": "Default_VRF_Universal"
            }
        ]
    }
}

```

**expected result**
vrfStatus of response is included in query result

mikewiebe:
Resolved by https://github.com/CiscoDevNet/ansible-dcnm/pull/22

PR #17, opened 2020-09-23
Dcnm phase1 enhancements

kharicha:
status of vrf should be included when query vrf - https://github.com/CiscoDevNet/ansible-dcnm/issues/15
status of network should be included when query - https://github.com/CiscoDevNet/ansible-dcnm/issues/16

Fixed the above and the respective new IT cases have been added for both.
Fixed the deploying fabric rest api call and the assert rules for it in all the IT cases in the module inventory/vrf/network.


tests/integration/targets/dcnm_vrf/tests/dcnm/query.yaml
```diff
@@ -90,6 +90,23 @@
...
+    fabric: "{{ ansible_it_fabric }}"
+    state: query
+  register: result
+
+- assert:
+    that:
+    - 'result.changed == false'
+    - 'result.response[0].fabric == "{{ ansible_it_fabric }}"'
+    - 'result.response[0].vrfName == "ansible-vrf-int1"'
+    - 'result.response[0].vrfStatus == "DEPLOYED"'
```
mikewiebe, line 108: Does the `vrfStatus` get returned if you query WITH the `config:` element?  Same question for network.

mikewiebe: It should if it does not.

kharicha: No, the status for both vrf and network comes only when we enquire about "all" the vrf/network present if specified with a specific vrf/network name the status is not populated.

mikewiebe: Why is that?  Is there no way to get the status when a user only wants to know about a specific set of vrfs or networks?  Seems like this is something we should support.

PR #18, opened 2020-09-23
Dcnm phase1 enhancements

kharicha:
status of vrf should be included when query vrf - #15
status of network should be included when query - #16

Fixed the above and the respective new IT cases have been added for both.
Fixed the deploying fabric rest api call and the assert rules for it in all the IT cases in the module inventory/vrf/network.

### DIFF
--- a/plugins/modules/dcnm_network.py
+++ b/plugins/modules/dcnm_network.py
@@ -1226,63 +1226,85 @@ class DcnmNetwork:
 
         query = []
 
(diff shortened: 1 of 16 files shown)


mikewiebe:
@kharicha Can we close this now since you are rewriting the query logic in the other PR?

mikewiebe:
These changes will not be committed.  Closing this since diffs will be stored in this closed PR for future reference.

PR #19, opened 2020-11-18
DCNM PHASE2 - P1 - VRF_LITE EXTENSION MODULE - Initial Commit

kharicha:
DCNM PHASE2 - P1 - VRF_LITE EXTENSION MODULE - Initial Commit
This is the PR raised for the following development in VRF LITE Extension module,
dcnm_vrf.py/IT modules/UT modules.

### DIFF
--- a/plugins/modules/dcnm_vrf.py
+++ b/plugins/modules/dcnm_vrf.py
@@ -1644,6 +1644,7 @@ def main():
         config=dict(required=False, type='list'),
         state=dict(default='merged',
                    choices=['merged', 'replaced', 'deleted', 'overridden', 'query'])
+        check_mode = dict(required=False, type="bool", default=False)
     )
 
     module = AnsibleModule(argument_spec=element_spec,
@@ -1684,7 +1685,8 @@ def main():
     else:
         module.exit_json(**dcnm_vrf.result)
 
-    if module.check_mode:
+    if module.params['check_mode']:
+        dcnm_vrf.result['changed'] = False
         module.exit_json(**dcnm_vrf.result)
 
     dcnm_vrf.push_to_remote()
--- a/plugins/modules/dcnm_vrf.py
+++ b/plugins/modules/dcnm_vrf.py
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import json
+import json, datetime
 import time
 import copy
 from ansible_collections.cisco.dcnm.plugins.module_utils.network.dcnm.dcnm import \
@@ -22,7 +22,7 @@ from ansible_collections.cisco.dcnm.plugins.module_utils.network.dcnm.dcnm impor
     dcnm_get_ip_addr_info, get_ip_sn_dict
 from ansible.module_utils.basic import AnsibleModule
 
-__author__ = "Shrishail Kariyappanavar"
+__author__ = "Shrishail Kariyappanavar, Karthik Babu Harichandra Babu"
 
 DOCUMENTATION = '''
 ---
@@ -30,8 +30,8 @@ module: dcnm_vrf
 short_description: Add and remove VRFs from a DCNM managed VXLAN fabric.
 version_added: "0.9.0"
 description:
-    - "Add and remove VRFs from a DCNM managed VXLAN fabric."
-author: Shrishail Kariyappanavar(@nkshrishail)
+    - "Add and remove VRFs and VRF Lite Extension from a DCNM managed VXLAN fabric."
+author: Karthik Babu Harichandra Babu (kharicha)
 options:
   fabric:
     description:
@@ -90,6 +90,37 @@ options:
             description: 'IP address of the switch where vrf will be attached or detached'
             type: ipv4
             required: true
+            suboptions:
+            vrf_lite:
+                description: VRF Lite Extensions options'
+                - interface: 
+                    description: 'Interface of the switch which is connected to the edge router'
+                    type: str
+                    requited: optional
+                  ipv4_addr:
+                    description: 'IP address of the interface which is connected to the edge router'
+                    type: ipv4
+                    requited: optional
+                  neighbor_ipv4:
+                    description: 'Neighbor IP address of the edge router'
+                    type: ipv4
+                    requited: optional
+                  ipv6_addr:
+                    description: 'IPv6 address of the interface which is connected to the edge router'
+                    type: ipv6
+                    requited: optional
+                  neighbor_ipv6:
+                    description: 'Neighbor IPv6 address of the edge router'
+                    type: ipv6
+                    requited: optional
+                  dot1q:
+                    description: 'DOT1Q Id'
+                    type: str
+                    requited: optional
+                  peer_vrf: 
+                    description: 'VRF Name to which this extension is attached'
+                    type: str
+                    requited: mandatory
           deploy:
             description: 'Per switch knob to control whether to deploy the attachment'
             type: bool
@@ -173,6 +204,31 @@ If rollback fails, the module does not attempt to rollback again, it just quits 
       attach:
       - ip_address: 192.168.1.224
       - ip_address: 192.168.1.225
+
+# VRF LITE Extension attached
+- name: Merge vrfs
+  cisco.dcnm.dcnm_vrf:
+    fabric: vxlan-fabric
+    state: merged
+    config:
+    - vrf_name: ansible-vrf-r1
+      vrf_id: 9008011
+      vrf_template: Default_VRF_Universal
+      vrf_extension_template: Default_VRF_Extension_Universal
+      vlan_id: 2000
+      source: null
+      service_vrf_template: null
+      attach:
+      - ip_address: 192.168.1.224
+      - ip_address: 192.168.1.225
+        vrf_lite:
+          - interface: Ethernet1/16 # optional
+            ipv4_addr: 10.33.0.2/30 # optional
+            neighbor_ipv4: 10.33.0.1 # optional
+            ipv6_addr: 2010::10:34:0:7/64 # optional
+            neighbor_ipv6: 2010::10:34:0:3 # optional
+            dot1q: 2 # dot1q can be got from dcnm/optional
+            peer_vrf: test_vrf_1 # peer_vrf is mandatory
 
 # The two VRFs below will be replaced in the target fabric.
 - name: Replace vrfs
@@ -284,6 +340,7 @@ If rollback fails, the module does not attempt to rollback again, it just quits 
       service_vrf_template: null
 '''
 
+
 class DcnmVrf:
 
     def __init__(self, module):
@@ -293,6 +350,9 @@ class DcnmVrf:
         self.fabric = module.params['fabric']
         self.config = copy.deepcopy(module.params.get('config'))
         self.check_mode = False
+        self.vrf_ext = False
+        self.role = ''
+        self.serial = ''
         self.have_create = []
         self.want_create = []
         self.diff_create = []
@@ -304,6 +364,7 @@ class DcnmVrf:
         self.have_attach = []
         self.want_attach = []
         self.diff_attach = []
+        self.validated   = []
         # diff_detach is to list all attachments of a vrf being deleted, especially for state: OVERRIDDEN
         # The diff_detach and delete operations have to happen before create+attach+deploy for vrfs being created.
         # This is specifically to address cases where VLAN from a vrf which is being deleted is used for another
@@ -315,6 +376,7 @@ class DcnmVrf:
         self.diff_deploy = {}
         self.diff_undeploy = {}
         self.diff_delete = {}
+        self.vrflitevalues = {}
         self.diff_input_format = []
         self.query = []
         self.inventory_data = get_fabric_inventory_details(self.module, self.fabric)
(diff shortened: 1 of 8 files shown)


tests/integration/targets/dcnm_vrf/tests/dcnm/merged.yaml
```diff
@@ -115,6 +115,107 @@
...
+- name: MERGED - Clean up any existing vrfs
+  cisco.dcnm.dcnm_vrf:
+    fabric: "{{ ansible_it_fabric }}"
+    state: deleted
+
+- name: MERGED - sleep for 10 seconds for DCNM to completely update the state
+  wait_for:
+    timeout: 10
+
+- name: MERGED - Create, Attach and Deploy new VRF - VLAN/VRF LITE EXTENSION Provided by the User in one switch
```
mikewiebe, line 127: I tried to run these VRF LITE tests but was not clear on what the topology should look like.  Can you share the topology you were using to run these tests?

Also, it looks like `Ethernet1/16` was significant for your topology but we need to think about making that interface a variable so that it can be passed in just like the switch ip and fabric name.

kharicha: The topology is nothing but 2 fabric, 1 external fabric and 1 normal fabric, 2 back to back connected devices each in one of the fabrics. E1/16 is not that significant in our code as that is the connected interface between these 2 devices in the topology and it is detected automatically by the dcnm. The interface is not at all mandatory for VRF LITE as we will pick that info from the dcnm only if not specified in the playbook.

plugins/modules/dcnm_vrf.py
```diff
@@ -342,15 +404,40 @@ def diff_for_attach_deploy(self, want_a, have_a):
             if have_a:
                 for have in have_a:
                     if want['serialNumber'] == have['serialNumber']:
-                        found = True
+                        if want['extensionValues'] != "" and have['extensionValues'] != "":
+                            want_ext_values = want['extensionValues']
+                            want_ext_values = eval(want_ext_values)
```
mikewiebe, line 409: Why are you using `eval` here?  There are some inherent security risks for eval so I think we should look for a different approach.

See: 
* https://nedbatchelder.com/blog/201206/eval_really_is_dangerous.html
* https://realpython.com/python-eval-function/#minimizing-the-security-issues-of-eval

plugins/modules/dcnm_vrf.py
```diff
@@ -342,15 +404,40 @@ def diff_for_attach_deploy(self, want_a, have_a):
...
+                            have_ext_values = have['extensionValues']
+                            have_ext_values = eval(have_ext_values)
+
+                            want_e = eval(want_ext_values['VRF_LITE_CONN'])
+                            have_e = eval(have_ext_values['VRF_LITE_CONN'])
+
+                            if want_e['VRF_LITE_CONN'][0]['IF_NAME'] == have_e['VRF_LITE_CONN'][0]['IF_NAME']:
```
mikewiebe, line 416: Given that both `want_e[key]0[key]` and `have_e[key]0[key]` have the same keys, we should be able to significantly simplify this logic and just do a dict compare:

For example:

```python
>>> want_e
{'VRF_LITE_CONN': [{'DOT1Q_ID': '429', 'IP_MASK': '255.255.255.0', 'NEIGHBOR_IP': '192.168.5.5'}]}
>>> have_e
{'VRF_LITE_CONN': [{'DOT1Q_ID': '429', 'IP_MASK': '255.255.255.0', 'NEIGHBOR_IP': '192.168.5.5'}]}
>>> 
>>> want_e == have_e
True
>>> 
>>> have_e['VRF_LITE_CONN'][0]['DOT1Q_ID'] = '500'
>>> 
>>> want_e
{'VRF_LITE_CONN': [{'DOT1Q_ID': '429', 'IP_MASK': '255.255.255.0', 'NEIGHBOR_IP': '192.168.5.5'}]}
>>> have_e
{'VRF_LITE_CONN': [{'DOT1Q_ID': '500', 'IP_MASK': '255.255.255.0', 'NEIGHBOR_IP': '192.168.5.5'}]}
>>> 
>>> want_e == have_e
False
>>> 
```
>>> want_e
{'VRF_LITE_CONN': [{'DOT1Q_ID': '429', 'IP_MASK': '255.255.255.0', 'NEIGHBOR_IP': '192.168.5.5'}]}
>>> have_e
{'VRF_LITE_CONN': [{'DOT1Q_ID': '429', 'IP_MASK': '255.255.255.0', 'NEIGHBOR_IP': '192.168.5.5', 'EXTRA_KEY': 'foo'}]}
>>> want_e == have_e
False
>>> 


This won't work for complex dict compares but in this case it should.

plugins/modules/dcnm_vrf.py
```diff
@@ -369,24 +456,73 @@ def update_attach_params(self, attach, vrf_name, deploy, vlanId=''):
...
+        self.role = role
+
         if role.lower() == 'spine' or role.lower() == 'super spine':
             msg = 'VRFs cannot be attached to switch {} with role {}'.format(attach['ip_address'], role)
             self.module.fail_json(msg=msg)
 
+        ext_values = {}
+        if attach['vrf_lite']:
+            '''Before apply the vrf_lite config, need double check if the swtich role is started wth Border'''
+            if (role.lower() != 'border' and
```
mikewiebe, line 475: Can you simplify by using a regular expression here and just check for `boarder` in the role name?

```python
>>> if not re.search(r'border', role.lower()):
...     print('VRF LITE cannot be attached')
... 
>>> role = 'spine'
>>> 
>>> if not re.search(r'border', role.lower()):
...     print('VRF LITE cannot be attached')
... 
VRF LITE cannot be attached
>>> 
```

plugins/modules/dcnm_vrf.py
```diff
@@ -369,24 +456,73 @@ def update_attach_params(self, attach, vrf_name, deploy, vlanId=''):
...
         attach.update({'vrfName': vrf_name})
         attach.update({'vlan': vlanId})
         attach.update({'deployment': deploy})
         attach.update({'isAttached': deploy})
         attach.update({'serialNumber': serial})
-        attach.update({'extensionValues': ""})
-        attach.update({'instanceValues': ""})
+        if self.vrf_ext:
+            attach.update({'extensionValues': json.dumps(ext_values).replace(' ','')})
+            attach.update({'instanceValues': "{\"loopbackId\":\"\",\"loopbackIpAddress\":\"\",\"loopbackIpV6Address\":\"\"}"})
```
mikewiebe, line 521: Can you use single quotes here so you don't have to escape `\` the double quotes `\"` ?

plugins/modules/dcnm_vrf.py
```diff
@@ -595,12 +774,16 @@ def get_want(self):
...
             return
 
-        for vrf in self.config:
+        for vrf in self.validated:
             vrf_attach = {}
             vrfs = []
 
             vrf_deploy = vrf.get('deploy', True)
-            vlanId = vrf.get('vlan_id', 0)
+            #vlanId = vrf.get('vlan_id', 0)
```
mikewiebe, line 782: Is this supposed to be commented out?

plugins/modules/dcnm_vrf.py
```diff
@@ -1173,8 +1363,92 @@ def push_to_remote(self, is_rollback=False):
...
 
-        method = 'POST'
         if self.diff_attach:
+            for d_a in self.diff_attach:
+                for v_a in d_a['lanAttachList']:
+                    if v_a.get('vrf_lite'):
+
+                        '''Before apply the vrf_lite config, need double check if the switch role is started wth Border'''
+
+                        if (self.role.lower() != 'border' and
```
mikewiebe, line 1373: Same comment here about using a regular expression to check for role `border` in one line instead of lines 1373-1378

plugins/modules/dcnm_vrf.py
```diff
@@ -1404,4 +1697,5 @@ def main():
 
 
 if __name__ == '__main__':
-    main()
\ No newline at end of file
+    main()
```
mikewiebe, line 1700: Looks like this change was unintended?

tests/integration/targets/dcnm_vrf/tests/dcnm/merged.yaml
mikewiebe: Understood, but when we run these tests it could be a different interface name that connects the fabrics right?  Just want to make sure we can pass whatever int is needed to the test instead of hard coding it.

plugins/modules/dcnm_vrf.py
```diff
@@ -90,6 +90,37 @@
             description: 'IP address of the switch where vrf will be attached or detached'
             type: ipv4
             required: true
+            suboptions:
+            vrf_lite:
+                description: VRF Lite Extensions options'
```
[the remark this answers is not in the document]
mikewiebe: Also below..

plugins/modules/dcnm_vrf.py
```diff
@@ -14,24 +14,24 @@
...
 DOCUMENTATION = '''
 ---
 module: dcnm_vrf
 short_description: Add and remove VRFs from a DCNM managed VXLAN fabric.
 version_added: "0.9.0"
 description:
-    - "Add and remove VRFs from a DCNM managed VXLAN fabric."
-author: Shrishail Kariyappanavar(@nkshrishail)
+    - "Add and remove VRFs and VRF Lite Extension from a DCNM managed VXLAN fabric."
+author: Karthik Babu Harichandra Babu (kharicha)
```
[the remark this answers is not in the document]
kharicha: added shrishail (but the complete code is almost changed with all the fixes and enhancements that already went in and now with vrf lite coming in which all was done by me :) )

plugins/modules/dcnm_vrf.py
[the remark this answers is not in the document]
kharicha: No i missed the leading ', added it now.

plugins/modules/dcnm_vrf.py
```diff
@@ -90,6 +90,37 @@
             description: 'IP address of the switch where vrf will be attached or detached'
             type: ipv4
             required: true
+            suboptions:
+            vrf_lite:
+                description: VRF Lite Extensions options'
+                - interface: 
+                    description: 'Interface of the switch which is connected to the edge router'
```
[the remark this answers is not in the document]
kharicha: This interface is part of the border router which connects to the edge routers in the external fabric.

plugins/modules/dcnm_vrf.py
```diff
@@ -14,24 +14,24 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import json
+import json, datetime
```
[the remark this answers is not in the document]
kharicha: removed.

plugins/modules/dcnm_vrf.py
```diff
@@ -90,6 +90,37 @@
...
+                    requited: optional
+                  neighbor_ipv6:
+                    description: 'Neighbor IPv6 address of the edge router'
+                    type: ipv6
+                    requited: optional
+                  dot1q:
+                    description: 'DOT1Q Id'
+                    type: str
+                    requited: optional
+                  peer_vrf:
```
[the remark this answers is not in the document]
kharicha: Done.

plugins/modules/dcnm_vrf.py
```diff
@@ -90,6 +90,37 @@
...
+                    requited: optional
+                  ipv4_addr:
+                    description: 'IP address of the interface which is connected to the edge router'
+                    type: ipv4
+                    requited: optional
+                  neighbor_ipv4:
+                    description: 'Neighbor IP address of the edge router'
+                    type: ipv4
+                    requited: optional
+                  ipv6_addr:
```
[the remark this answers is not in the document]
kharicha: Yes i added both the ipv4 and ipv6 at the same time in my UT/IT tests.

plugins/modules/dcnm_vrf.py
```diff
@@ -174,6 +205,31 @@
...
+      - ip_address: 192.168.1.224
+      - ip_address: 192.168.1.225
+        vrf_lite:
+          - interface: Ethernet1/16 # optional
+            ipv4_addr: 10.33.0.2/30 # optional
+            neighbor_ipv4: 10.33.0.1 # optional
+            ipv6_addr: 2010::10:34:0:7/64 # optional
+            neighbor_ipv6: 2010::10:34:0:3 # optional
+            dot1q: 2 # dot1q can be got from dcnm/optional
+            peer_vrf: test_vrf_1 # peer_vrf is mandatory
```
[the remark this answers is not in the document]
kharicha: Done.

plugins/modules/dcnm_vrf.py
kharicha: I am using it for converting pupose. Will need to look for better approach, may be we can discuss.

plugins/modules/dcnm_vrf.py
kharicha: Done.

plugins/modules/dcnm_vrf.py
kharicha: yes it needs to be, removed the same.

plugins/modules/dcnm_vrf.py
kharicha: done.

plugins/modules/dcnm_vrf.py
kharicha: yes removed the extra line at the last.

tests/integration/targets/dcnm_vrf/tests/dcnm/merged.yaml
kharicha: ok will change the interface details to variable.

kharicha: done the same in IT cases.

plugins/modules/dcnm_vrf.py
```diff
@@ -174,6 +205,31 @@
...
+      attach:
+      - ip_address: 192.168.1.224
+      - ip_address: 192.168.1.225
+        vrf_lite:
+          - interface: Ethernet1/16 # optional
+            ipv4_addr: 10.33.0.2/30 # optional
+            neighbor_ipv4: 10.33.0.1 # optional
+            ipv6_addr: 2010::10:34:0:7/64 # optional
+            neighbor_ipv6: 2010::10:34:0:3 # optional
+            dot1q: 2 # dot1q can be got from dcnm/optional
```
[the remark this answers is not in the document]
kharicha: added.

plugins/modules/dcnm_vrf.py
kharicha: done.

plugins/modules/dcnm_vrf.py
kharicha: Tried this for all the IT cases, i see some idempotence cases are failing due to this change, reverted to the original.

plugins/modules/dcnm_vrf.py
kharicha: We are just adding this line straight from how the DCNM displays the instance values, we are not adding or verifying the same anywhere, hence planning to leave it as it is.

plugins/modules/dcnm_vrf.py
mikewiebe: This is still something we need to change.  We should not use eval.

Possible alternative:  https://eulertech.wordpress.com/2018/06/10/the-pitfall-of-eval-function-and-its-safe-alternative-in-python/

plugins/modules/dcnm_vrf.py
mikewiebe: Ok for now.

plugins/modules/dcnm_vrf.py
```diff
@@ -293,6 +356,9 @@ def __init__(self, module):
         self.fabric = module.params['fabric']
         self.config = copy.deepcopy(module.params.get('config'))
         self.check_mode = False
```
mikewiebe, line 358: Not really in scope for this PR but do we have tests for check_mode?

kharicha: i havent done it before, will discuss with you on the call.

plugins/modules/dcnm_vrf.py
kharicha: done, changed it with ast.literal_eval.

kharicha: Tested the code with UT and IT after the eval change to ast.literal_eval, all success.

plugins/modules/dcnm_vrf.py
kharicha: Added the check mode cases in ut/it and verified the results.

PR #20, opened 2020-11-24
DCNM policy and template modules initial commit after UT andf IT

mmudigon:


plugins/modules/dcnm_policy.py
```diff
@@ -0,0 +1,1072 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+__author__ = "Mallik Mudigonda"
+
+DOCUMENTATION = '''
+---
+module: dcnm_policy
+short_description: DCNM Ansible Module for managing policies.
+version_added: "0.9.0"
```
mikewiebe, line 23: This needs to be updated to `1.1.0`

plugins/modules/dcnm_policy.py
```diff
@@ -0,0 +1,1072 @@
...
+    """
+    element_spec = dict(
+        fabric=dict(required=True, type='str'),
+        config=dict(required=False, type='list'),
+        state=dict(type='str', default='merged',
+                   choices = ['merged', 'deleted', 'query']),
+        deploy=dict(required=False, type='bool', default=True)
+    )
+
+    with open ("check.log", "w+") as f:
```
mikewiebe, line 1010: This write to file should be removed for the final version right?

plugins/modules/dcnm_policy.py
```diff
@@ -0,0 +1,1072 @@
...
+    
+    start = datetime.datetime.now()
+
+    if not dcnm_policy.ip_sn:
+        dcnm_policy.result['msg'] = "Fabric {} missing on DCNM or does not have any switches".format(dcnm_policy.fabric)
+        module.fail_json(msg="Fabric {} missing on DCNM or does not have any switches".format(dcnm_policy.fabric))
+
+    state = module.params['state']
+    if not dcnm_policy.config:
+        if state == 'merged' or state == 'deleted' or state == 'query':
```
mikewiebe, line 1029: I can't remember what we discussed.  Did we decide that `config` was mandatory for `deleted` state?  This means we cannot omit `config` and in order to delete all policies?

plugins/modules/dcnm_template.py
```diff
@@ -0,0 +1,629 @@
...
+        self.fd            = None
+        self.changed_dict  = [{'merged' : [], 'deleted' : [], 'query' : [], 'failed' : []}]
+
+        self.result = dict(
+            changed  = False,
+            diff     = [],
+            response = []
+        )
+
+    def log_msg (self, msg):
```
mikewiebe, line 197: Same question here for `log_msg`

plugins/modules/dcnm_template.py
```diff
@@ -0,0 +1,629 @@
...
+                h = re.sub (' +', '', have['content']).rstrip('\n')
+
+                if (w != h):
+                    return 'DCNM_TEMPLATE_MERGE'
+                else: 
+                    return 'DCNM_TEMPLATE_DONT_ADD'
+
+        return 'DCNM_TEMPLATE_ADD_NEW'
+
+    def dcnm_template_get_have_all (self):
```
mikewiebe, line 317: Is this method needed?

tests/integration/targets/dcnm_policy/tests/dcnm/dcnm_policy_merge.yaml
```diff
@@ -0,0 +1,516 @@
...
+  loop: '{{ result.response }}'
+
+
+- block:
+
+##############################################
+##                MERGE                     ##
+##############################################
+
+    - name: Create different non-existing policies
```
mikewiebe, line 31: Just want to make sure I am reading this task correctly.

This will do the following:

* Create all policies on switch1
* If the IP on line 41 was changed to switch2
    * 104 and 105 would only be created on switch2
    * Would 101, 102 and 103 also be created on switch2 or would the policies under switch2 act as an override for the global policies?

mikewiebe (changes requested):

plugins/modules/dcnm_policy.py
mmudigon: Updated the code

plugins/modules/dcnm_policy.py
```diff
@@ -0,0 +1,1072 @@
...
+__author__ = "Mallik Mudigonda"
+
+DOCUMENTATION = '''
+---
+module: dcnm_policy
+short_description: DCNM Ansible Module for managing policies.
+version_added: "0.9.0"
+description:
+    - "DCNM Ansible Module for Creating, Deleting, Querying and Modifying
+    - policies
```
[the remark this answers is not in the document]
mmudigon: Updated

plugins/modules/dcnm_policy.py
```diff
@@ -0,0 +1,1072 @@
...
+                default: {}
+'''
+
+EXAMPLES = '''
+
+States:
+This module supports the following states:
+
+Merged:
+  Interfaces defined in the playbook will be merged into the target fabric.
```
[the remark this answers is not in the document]
mmudigon: Updated

plugins/modules/dcnm_policy.py
```diff
@@ -0,0 +1,1072 @@
...
+
+  The policies listed in the playbook will be created if not already present on the DCNM
+  server. If the policy is already present and the configuration information included
+  in the playbook is either different or not present in DCNM, then the corresponding
+  information is added to the policy on DCNM. If an policy mentioned in playbook
+  is already present on DCNM and there is no difference in configuration, no operation
+  will be performed for such policy.
+
+Deleted:
+  Interfaces defined in the playbook will be deleted in the target fabric.
```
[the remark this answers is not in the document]
mmudigon: Updated

plugins/modules/dcnm_policy.py
```diff
@@ -0,0 +1,1072 @@
...
+
+      - switch:
+          - ip: "{{ ansible_switch1 }}"
+            policies:
+              - name: template_104  # This must be a valid template name
+                create_additional_policy: false  # Do not create a policy if it already exists
+
+              - name: template_105  # This must be a valid template name
+                create_additional_policy: false  # Do not create a policy if it already exists
+              - ip: "{{ ansible_switch2 }}"
```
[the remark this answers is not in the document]
mmudigon: No, that was not right. Corrected the indentation

plugins/modules/dcnm_policy.py
mmudigon: Removed

plugins/modules/dcnm_policy.py
mmudigon: Yes. Because there are many policies on the server and we cannot delete all. So we decided that the user must explicitly state which policies need to be deleted

plugins/modules/dcnm_template.py
mmudigon: I just left the function definition alone, in case we need this for debugging later. This will not be called in the final version.

plugins/modules/dcnm_template.py
mmudigon: No. not required. Removed

tests/integration/targets/dcnm_policy/tests/dcnm/dcnm_policy_merge.yaml
mmudigon: Yes, you are right. If policies are mentioned under a switch, then they will override all others. In this case as you said 104 and 105 will be created on switch2 and the other policies on switch1. But since I had a single switch I used switch1 again. But your understanding is right

tests/integration/targets/dcnm_template/tests/dcnm/dcnm_template_nodel.yaml
```diff
@@ -0,0 +1,92 @@
+##############################################
+##               SETUP                      ##
```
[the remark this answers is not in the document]
mmudigon: Sorry for the confusing nomenclature.  This file includes test cases where we try to delete templates which are used by policies. So delete should not happen. I will change the name to dcnm_template_no_delete.yaml

plugins/modules/dcnm_template.py
```diff
@@ -0,0 +1,629 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+__author__ = "Mallik Mudigonda"
+
+DOCUMENTATION = '''
+---
+module: dcnm_template
+short_description: DCNM Ansible Module for managing templates.
+version_added: "1.0.0"
```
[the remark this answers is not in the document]
mmudigon: Updated

plugins/modules/dcnm_policy.py
```diff
@@ -0,0 +1,1072 @@
...
+        self.inventory_data = get_fabric_inventory_details(self.module, self.fabric)
+        self.ip_sn, self.hn_sn = get_ip_sn_dict(self.inventory_data)
+
+        self.result = dict(
+            changed  = False,
+            diff     = [],
+            response = []
+        )
+
+    def log_msg (self, msg):
```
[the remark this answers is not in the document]
mmudigon: Intentional. Left it so that we can use it for debugging laster.

plugins/modules/dcnm_policy.py
```diff
@@ -0,0 +1,1072 @@
...
+            self.pb_input.append(copy.deepcopy(cfg))
+
+    def dcnm_policy_validate_input (self):
+
+        policy_spec = dict(
+            name                      = dict(required=True, type='str'),
+            create_additional_policy  = dict(required=False, type='bool', default=True),
+            description               = dict(required=False, type='str', default =''),
+            priority                  = dict(required=False, type=int, default=500),
+            Vars                      = dict(required=False, type=dict, default={}),
```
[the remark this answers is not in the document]
mmudigon: Yes. I also felt that way. Wanted to check with you. Will update to policy_vars

plugins/modules/dcnm_policy.py
```diff
@@ -0,0 +1,1072 @@
...
+
+      switch:
+        description:
+          - A dictionary of switches and associated policy information. All switches in this list will be deployed with only those policies
+            that are included under "policies" object i.e. 'policies' object will override the list of policies for this particular switch.
+            If 'policies' object is not included, then other policies specified in the configurstion will be deployed to these switches.
+        type: list
+        elements: dict
+        suboptions:
+          ip: IP address of the switch where the policy is to be deployed
```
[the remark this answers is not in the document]
mmudigon: Changed it to string since we can have hostnames also

plugins/modules/dcnm_policy.py
```diff
@@ -0,0 +1,1072 @@
...
+class DcnmPolicy:
+
+    def __init__(self, module):
+        self.module         = module
+        self.params         = module.params
+        self.fabric         = module.params['fabric']
+        self.config         = copy.deepcopy(module.params.get('config'))
+        self.deploy         = True # Global 'deploy' flag
+        self.pb_input       = []
+        self.check_mode     = False
```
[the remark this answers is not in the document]
mmudigon: Done

PR #22, opened 2020-12-07
DCNM VRF AND NETWORK - QUERY REWORK - IT/UT

kharicha:
DCNM VRF AND NETWORK - QUERY REWORK - IT/UT


tests/integration/targets/dcnm_network/tests/dcnm/merged.yaml
```diff
@@ -249,6 +249,55 @@
   wait_for:
     timeout: 20
 
+- name: MERGED - Create, Attach and Deploy Multiple Network with Single Switch Attach wuth check mode set to true
```
mikewiebe, line 252: `/wuth/with`

tests/integration/targets/dcnm_network/tests/dcnm/query.yaml
```diff
@@ -103,20 +103,60 @@
...
-    - 'result.response[0].vlan_id == "1500"'
-    - 'result.response[0].vrf_name == "Tenant-1"'
-    - 'result.response[1].gw_ip_subnet == "192.168.40.1/24"'
-    - 'result.response[1].net_extension_template == "Default_Network_Extension_Universal"'
-    - 'result.response[1].net_id == 7002'
-    - 'result.response[1].net_name == "ansible-net12"'
-    - 'result.response[1].net_template == "Default_Network_Universal"'
-    - 'result.response[1].vlan_id == "151"'
-    - 'result.response[1].vrf_name == "Tenant-2"'
+    - 'result.response[0].parent.networkName == "ansible-net13"'
```
mikewiebe, line 106: The structure that gets returned for query is quite different the the old structure and I no longer see some of the information in the test that we had previously.  What about `gw_ip_subnet`?  Do we still get that information in the query?

If we have users that are already using query but the structure of the dictionary has changed we might break existing playbooks.  Can we return the same structure and key names using the new query logic?

mikewiebe (changes requested):

tests/integration/targets/dcnm_network/tests/dcnm/query.yaml
kharicha: We have not changed anything from the respone point of view, we are infact giving more response as part of the new logic.
The response contains the gw_ip_subent but this is inside a list of list of dict, i dont want to go inside the nested list for getting those values for the query IT cases, any how those fields have been tested and checked extensively in other state UT cases.

tests/integration/targets/dcnm_network/tests/dcnm/merged.yaml
kharicha: changed.

tests/integration/targets/dcnm_vrf/tests/dcnm/query.yaml
```diff
@@ -82,13 +82,17 @@
 - assert:
     that:
     - 'result.changed == false'
-    - 'result.response[0].attach[0].deploy == true'
```
mikewiebe, line 85: @kharicha Quick question here.  Now that you have added the old fields back we can add these old tests back right?  Just want to make sure we have coverage for both old and new structure until we deprecate the old.

kharicha: yes we have added all the checks new ones and also the old ones,

- assert:
    that:
    - 'result.changed == true'
    - 'result.response[0].RETURN_CODE == 200'
    - 'result.response[1].RETURN_CODE == 200'
    - 'result.response[2].RETURN_CODE == 200'
    - '(result.response[1].DATA|dict2items)[0].value == "SUCCESS"'
    - '(result.response[1].DATA|dict2items)[1].value == "SUCCESS"'
    - 'result.diff[0].attach[0].deploy == true'
    - 'result.diff[0].attach[1].deploy == true'
    - '"{{ ansible_switch1 }}" in result.diff[0].attach[0].ip_address'
    - '"{{ ansible_switch2 }}" in result.diff[0].attach[1].ip_address'
    - 'result.diff[0].vrf_name == "ansible-vrf-int2"'

we can see that deploy from the old ones without parent/attach output is also checked.

PR #29, opened 2020-12-18
Add sample hosts file

mikewiebe:
We have received several requests to add a sample hosts file for authenticating the connection to DCNM.  This PR adds a sample hosts file to the README in both INI and YAML formats to clarify how the httpapi dcnm plugin is used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ The following example task adds a switch to an existing fabric, using the FQCN:
 ```yaml
 ---
 
-- hosts: dcnm
+- hosts: dcnm_controllers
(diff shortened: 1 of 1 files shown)


mmudigon (approved):
The content looks good. But what is the specific change you were referring to as 'authentication' in the mail?

mikewiebe:
> The content looks good. But what is the specific change you were referring to as 'authentication' in the mail?

@mmudigon I was simply referring to the format for specifying credentials used to log into DCNM using our httpapi dcnm connection plugin

PR #34, opened 2021-01-20
Fixes in template module to support args and also enable required feature for IT

mmudigon:
This is a bug fix in template module. The following are the fixes:

1. Template creation with argument support had an issue in the "content". That was updated
2. Since testing of policy module require support for telemetry feature on the box, the feature enablement is now done in the prepare_dcnm_policy module at the beginning of the tests.
3. Deploy response for templates with arguments is different from templates without arguments. Updated the code handling the response.

### DIFF
--- a/plugins/modules/dcnm_template.py
+++ b/plugins/modules/dcnm_template.py
@@ -242,7 +242,7 @@ class DcnmTemplate:
     def dcnm_template_get_template_payload(self, ditem):
 
         if self.module.params["state"] == "merged":
-            std_cont = "##template properties\nname = __TEMPLATE_NAME;\ndescription = __DESCRIPTION;\ntags = __TAGS;\nuserDefined = true;\nsupportedPlatforms = All;\ntemplateType = POLICY;\ntemplateSubType = DEVICE;\ncontentType = TEMPLATE_CLI;\nimplements = implements;\ndependencies = ;\npublished = false;\n##\n##template variables\n##\n##template content\n"
+            std_cont = "##template properties\nname = __TEMPLATE_NAME;\ndescription = __DESCRIPTION;\ntags = __TAGS;\nuserDefined = true;\nsupportedPlatforms = All;\ntemplateType = POLICY;\ntemplateSubType = DEVICE;\ncontentType = TEMPLATE_CLI;\nimplements = implements;\ndependencies = ;\npublished = false;\n##\n##template content\n"
 
             template_payload = {}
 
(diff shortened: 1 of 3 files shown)


plugins/modules/dcnm_template.py
```diff
@@ -334,11 +334,19 @@ def dcnm_template_validate_template(self, template):
...
-                if d["reportItemType"].lower() == "error":
+            # resp['DATA'] may be a list in case of templates with no parameters. But for templates
+            # with parameters resp['DATA'] will be a dict directly with 'status' as 'Template Validation Successful'
+            if isinstance(resp['DATA'], list):
+                for d in resp["DATA"]:
+                    if d.get("reportItemType", ' ').lower() == "error":
+                        self.result["response"].append(resp)
+                        return 0
+                return resp["RETURN_CODE"]
+            elif isinstance(resp['DATA'], dict):
```
mikewiebe, line 345: Should you have a final `else` block in the event that the `resp['DATA']` is something we don't recognize to generate a meaningful/helpful failure?

plugins/modules/dcnm_template.py
```diff
@@ -334,11 +334,19 @@ def dcnm_template_validate_template(self, template):
...
 
-            for d in resp["DATA"]:
-                if d["reportItemType"].lower() == "error":
+            # resp['DATA'] may be a list in case of templates with no parameters. But for templates
+            # with parameters resp['DATA'] will be a dict directly with 'status' as 'Template Validation Successful'
+            if isinstance(resp['DATA'], list):
+                for d in resp["DATA"]:
+                    if d.get("reportItemType", ' ').lower() == "error":
+                        self.result["response"].append(resp)
+                        return 0
```
mikewiebe, line 343: Why are you returning `0` in this case?  What is the `resp["RETURN_CODE"]` value when this condition is true?

plugins/modules/dcnm_template.py
```diff
@@ -334,11 +334,19 @@ def dcnm_template_validate_template(self, template):
...
+                return resp["RETURN_CODE"]
+            elif isinstance(resp['DATA'], dict):
+                if resp['DATA'].get("status",  ' ').lower() != "template validation successful":
                     self.result["response"].append(resp)
                     return 0
```
mikewiebe, line 348: Same question here about returning `0`

mmudigon:
Not very clean after I made further changes. so closing it and opening another after addressing all comments from this PR

PR #35, opened 2021-01-21
Add full fabric integration tests and github actions ci tasks

mikewiebe:
Summary of updates:

* Add new gihub actions CI file to build collection, run unit tests and generate simple code coverage report.
* Add new dcnm_fabric role tests to validate modules using a full fabric
    * Additional coverage will be added to these tests.  This is only a starting point.
    * More tests to be added once dcnm_vrf and dcnm_network query capabilities added to query based on object key
* Fixes the following bugs with dcnm_vrf module
    * Type conversion for vrf config data when read from an ansible file
    * VRFs were deployed even when the deploy flag was set to false
    * Removed source parameter from examples and hardcoded it to None.  This should not be exposed in the playbook parameters but leaving it here just in case customers have specified it in their playbooks.
    * Whitespace issues

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,9 +6,9 @@ name: CI
 on:
   # Triggers the workflow on push or pull request events but only for the develop branch
   push:
-    branches: [ test-ci ]
+    branches: [ develop ]
   pull_request:
-    branches: [ test-ci ]
+    branches: [ develop ]
 
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,88 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
+
+# Controls when the action will run.
+on:
+  # Triggers the workflow on push or pull request events but only for the develop branch
+  push:
+    branches: [ test-ci ]
+  pull_request:
+    branches: [ test-ci ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    name: Build DCNM collection
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        ansible: [2.9.12, 2.10.0]
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v2
+    
+    - name: Set up Python 3.7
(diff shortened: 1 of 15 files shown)


.github/workflows/main.yml
```diff
@@ -0,0 +1,88 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
+
+# Controls when the action will run.
+on:
+  # Triggers the workflow on push or pull request events but only for the develop branch
+  push:
+    branches: [ test-ci ]
```
mmudigon, line 9: Do you want to change this to develop branch before commit or is it going to a reference commit on a test_ci branch

.github/workflows/main.yml
```diff
@@ -0,0 +1,88 @@
...
+    - name: Check out code
+      uses: actions/checkout@v2
+    
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+
+    - name: Install ansible-base (v${{ matrix.ansible }})
+      run: pip install https://github.com/ansible/ansible/archive/v${{ matrix.ansible }}.tar.gz --disable-pip-version-check
```
mmudigon, line 35: Sorry did not get this point. Why are we using pip install instead of git clone directly. Because this is how we do install on our local MAC. Is this because we want to get TGZ files for faster response

.github/workflows/main.yml
```diff
@@ -0,0 +1,88 @@
...
+    strategy:
+      matrix:
+        ansible: [2.9.12, 2.10.0]
+    steps:
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+
+    - name: Install ansible-base (v${{ matrix.ansible }})
```
mmudigon, line 60: Why install ansible-base again? We did that in the build.

.github/workflows/main.yml
mikewiebe: Good catch!  Will update it.

.github/workflows/main.yml
mikewiebe: Good question.  We could investigate using git clone in the future but for now I was following the same process from the ACI/MSO collection.  Not sure if this is faster but we just need a stable base ansible collection to be in place.

.github/workflows/main.yml
mikewiebe: This is another case where I used the ACI/MSO as a reference.  We *might* not need this step if the environment carries over from the build section but we will need to test to verify.  For now let's go ahead and get this committed as a starting point and then we can test futher optimizations in a future PR.

PR #36, opened 2021-01-22
DCNM VRF - Query filter fixed

kharicha:
Fixed the dcnm vrf - query filter based on vrf name/vrf id and vlan.
vrf name is mandatory in config and rest are options.


mikewiebe:
Close and reopen to restart CI tests.

PR #48, opened 2021-04-01
Changes for multisite support in vrf module

praveenramoorthy:
These changes are for supporting multisite fabrics in vrf module.


plugins/module_utils/network/dcnm/dcnm.py
```diff
@@ -152,6 +152,20 @@ def get_ip_sn_dict(inventory_data):
     return ip_sn, hn_sn
 
 
+def get_ip_sn_fabric_dict(inventory_data):
```
mikewiebe, line 155: Can you add a function doc string for this module to describe what it's doing.  Want to make sure we start doing this for all new functions we add as a best practice.

plugins/module_utils/network/dcnm/dcnm.py
```diff
@@ -262,6 +276,47 @@ def dcnm_get_ip_addr_info(module, sw_elem, ip_sn, hn_sn):
 #             raise Exception(json.dumps(msg_dict))
 
 
+def get_fabric_details(module, fabric):
```
mikewiebe, line 279: Can you add a function doc string for this module to describe what it's doing.  Want to make sure we start doing this for all new functions we add as a best practice.

mmudigon, line 294: WE already have a function to get FABRIC details. Is this something different? Otherwise use the existing one

plugins/modules/dcnm_vrf.py
```diff
@@ -429,10 +435,11 @@ def diff_for_attach_deploy(self, want_a, have_a):
                         else:
                             found = True
 
-                            if bool(have['isAttached']) is not bool(want['isAttached']):
-                                del want['isAttached']
-                                attach_list.append(want)
-                                continue
+                            if want.get('isAttached') is not None:
+                                if bool(have['isAttached']) is not bool(want['isAttached']):
```
mmudigon, line 439: bool(True) is True
bool('False') is also True

Not sure what you are checking here?

plugins/module_utils/network/dcnm/dcnm.py
praveenramoorthy: Added doc string

plugins/module_utils/network/dcnm/dcnm.py
praveenramoorthy: Added doc string

praveenramoorthy: The function which is there already gets the fabric inventory details. Here we get the fabric details

plugins/modules/dcnm_vrf.py
praveenramoorthy: There was this traceback when we try to delete a vrf which is not attached to any switch as the key 'isAttached' was not there in the want dict. want.get checks if the key is there in the dict, it returns the value if key is present, if not it returns none. The intention here is to continue with our check if 'isAttached' is there in dict, else we skip the check.

plugins/module_utils/network/dcnm/dcnm.py
```diff
@@ -262,6 +290,56 @@ def dcnm_get_ip_addr_info(module, sw_elem, ip_sn, hn_sn):
 #             raise Exception(json.dumps(msg_dict))
 
 
+# This call is used to get the details of the given fabric from the DCNM
+def get_fabric_details(module, fabric):
+    """
+    Used to get the details of the given fabric from the DCNM
+
+    Parameters:
```
mikewiebe, line 298: Looks like you are missing `module` as one of the parameters in this doc string

plugins/modules/dcnm_vrf.py
```diff
@@ -33,6 +33,8 @@
 version_added: "0.9.0"
 description:
     - "Add and remove VRFs and VRF Lite Extension from a DCNM managed VXLAN fabric."
+    - "In Multisite fabrics, VRFs can be created only on Multisite fabric"
```
mikewiebe, line 36: I can't remember, but if you attempt to create a vrf on the member fabric does DCNM generate a meaningful error message?

plugins/modules/dcnm_vrf.py
```diff
@@ -429,10 +435,11 @@ def diff_for_attach_deploy(self, want_a, have_a):
                         else:
                             found = True
 
-                            if bool(have['isAttached']) is not bool(want['isAttached']):
-                                del want['isAttached']
-                                attach_list.append(want)
-                                continue
+                            if want.get('isAttached') is not None:
```
mikewiebe, line 438: Please add a comment here describing why `want['isAttached']` is being deleted in this case

plugins/modules/dcnm_vrf.py
```diff
@@ -1431,59 +1463,65 @@ def push_to_remote(self, is_rollback=False):
...
+                                    ms_con = {}
+                                    ms_con['MULTISITE_CONN'] = []
+                                    extension_values['MULTISITE_CONN'] = json.dumps(ms_con)
+
+                                    v_a['extensionValues'] = json.dumps(extension_values).replace(' ', '')
+                                    v_a['instanceValues'] = "{\"loopbackId\":\"\",\"loopbackIpAddress\":\"\",\"loopbackIpV6Address\":\"\"}"
+                                    del v_a['vrf_lite']
+
+                        if ext_values == None:
+                            msg = 'There is no VRF LITE capable interface in switch {}'.format(attach['ip_address'])
```
mikewiebe, line 1523: *nit:* I would reword the `msg` slightly to read:

`'There is no VRF LITE capable interface on this switch {}'`

tests/integration/targets/dcnm_vrf/tests/dcnm/overridden.yaml
```diff
@@ -201,12 +201,8 @@
     - 'result.response[0].RETURN_CODE == 200'
     - 'result.response[1].RETURN_CODE == 200'
     - '(result.response[0].DATA|dict2items)[0].value == "SUCCESS"'
-    - '(result.response[0].DATA|dict2items)[1].value == "SUCCESS"'
```
mikewiebe, line 204: Remind me why we are deleting this check?

tests/integration/targets/dcnm_vrf/tests/dcnm/query.yaml
```diff
@@ -88,11 +88,9 @@
     - 'result.response[0].attach[0].switchDetailsList[0].islanAttached == true'
     - 'result.response[0].attach[0].switchDetailsList[0].lanAttachedState == "DEPLOYED"'
     - 'result.response[0].attach[0].switchDetailsList[0].vlan == 500'
-    - 'result.response[0].attach[0].switchDetailsList[0].extensionValues == ""'
```
mikewiebe, line 91: Why are we removing this check?

mikewiebe (changes requested):

mikewiebe:
@praveenramoorthy Just a few more minor comments

plugins/module_utils/network/dcnm/dcnm.py
praveenramoorthy: Added

plugins/modules/dcnm_vrf.py
praveenramoorthy: If we attempt to add vrf to a member site, we just return the error whatever DCNM returns. DCNM returns "Invalid Fabric" as message in this scenario. The fabric details we fetch for the member site, does not specify that it is a member site of a multisite fabric. On creation this is just another easy fabric, only on moving this fabric to a multisite fabric, it becomes a member fabric. So only the multisite fabric details have the information on corresponding member site fabrics. And in this particular case if the user tries to configure a vrf in member site, it will be difficult to determine if it is a member site of a multisite fabric to generate a custom message in corresponding Ansible module.

plugins/modules/dcnm_vrf.py
praveenramoorthy: Changed as suggested

tests/integration/targets/dcnm_vrf/tests/dcnm/query.yaml
praveenramoorthy: This check is being removed as in case of multisite fabrics, the extensionValues field will not be empty. It will have some values. So in order to make same playbook work for both normal and multisite fabric, this check has been removed.

plugins/modules/dcnm_vrf.py
praveenramoorthy: Added

tests/integration/targets/dcnm_vrf/tests/dcnm/overridden.yaml
praveenramoorthy: There was an issue in vrf module, in the following scenario. If a switch is already attached to a vrf and the same switch is being attached again along with another switch to same vrf, then the attach was sent to both the switches. The attach request was being sent for switch which is already attached, which ideally should not have happened. The IT cases were written to check for 2 attach in response. The attached should have been sent only for the new switch being added. This works fine for network module(attach is sent for newly added switch only). There was an issue in vrf module code while determining the switches which are already attached. Have fixed this issue now, and after the fix the attach is being sent only for the newly added switch and I have removed the check for second attach response in IT cases as it will not be there.

mikewiebe (approved):

Issue #63, opened 2021-04-24
Role[border_gateway]is not a valid role


dsx1123:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Ansible Version and collection version

```
ansible 2.10.5
  config file = /root/.ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /root/virtualenv/ansible/lib/python3.7/site-packages/ansible
  executable location = /root/virtualenv/ansible/bin/ansible
  python version = 3.7.3 (default, Apr  3 2019, 05:39:12) [GCC 8.3.0]
```
### DCNM version

* V 11.5.1

### Affected module(s)

* dcnm_inventory

### Ansible Playbook

```yaml
# Copy-paste your anisble playbook here 
---
- name: add switches to fabric
  hosts: dcnm
  gather_facts: false
  collections:
    - cisco.dcnm
  tasks:
    - name: add switches to fabric
      dcnm_inventory:
        fabric: "{{ lab109_fabric_1.fabric }}"
        config:
          - seed_ip: 172.31.186.152
            user_name: admin
            password: xxxxx
            max_hops: 0
            role: border_gateway
            preserve_config: false

```

### Debug Output
https://gist.github.com/dsx1123/153690db572ff4fff9e3ed52ace027e9

checked the API call, the role used in API is actually "border gateway", however if set the role to this string, it will not pass the validation. Only tested border gateway but I suspect same issue could apply to all other multi-word role.

### Expected Behavior

Switch should be added to fabric as border gateway

### Actual Behavior

see debug output 

### Steps to Reproduce

execute playbook above to register switch to exsited fabric

mikewiebe:
@dsx1123 Thanks for filing the issue.  We are working on a fix

PR #64, opened 2021-04-29
Fix for setting role of switch

praveenramoorthy:
This PR fixes fixes the issue in setting the role of switch in #63

### DIFF
--- a/plugins/modules/dcnm_inventory.py
+++ b/plugins/modules/dcnm_inventory.py
@@ -674,7 +674,7 @@ class DcnmInventory:
                     self.module.fail_json(msg=msg)
                 if role['ipAddress'] == create["switches"][0]['ipaddr']:
                     method = 'PUT'
-                    assign_path = '/fm/fmrest/topology/role/{}?newRole={}'.format(role['switchDbID'], create['role'])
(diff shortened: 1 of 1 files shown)


plugins/modules/dcnm_inventory.py
```diff
@@ -674,7 +674,7 @@ def assign_role(self):
                     self.module.fail_json(msg=msg)
                 if role['ipAddress'] == create["switches"][0]['ipaddr']:
                     method = 'PUT'
-                    assign_path = '/fm/fmrest/topology/role/{}?newRole={}'.format(role['switchDbID'], create['role'])
+                    assign_path = '/fm/fmrest/topology/role/{}?newRole={}'.format(role['switchDbID'], create['role'].replace("_", "%20"))
```
mikewiebe, line 677: Were you able to verify that all role choices not work properly with this fix?

praveenramoorthy: The issue was with the role choices having "_" in there name(border_gateway, border_gateway_super_spine, etc). I have verified all the role choices with "_" in their name along with spine and border, and it works properly as expected.

Issue #60, opened 2021-04-21
Ansible playbook for adding policy template with argument values does not get configured on DCNM


micruzz82:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Ansible Version and collection version

ansible 2.10.8
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.6/dist-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.6.9 (default, Jan 26 2021, 15:33:00) [GCC 8.4.0]

ansible-galaxy collection list | grep dcnm
cisco.dcnm        1.1.0 

### DCNM version

* V 11.5.1

### Affected module(s)

* cisco.dcnm.dcnm_policy

### Ansible Playbook

```yaml
---

- hosts: dcnm
  gather_facts: false
  connection: ansible.netcommon.httpapi

  vars:
    ansible_it_fabric: dcnm-fabric
    ansible_switch1: 10.20.20.1
    ansible_switch2: 10.20.20.2

  tasks:
    - name: Create policy including required variables
      cisco.dcnm.dcnm_policy:
        fabric: "{{ ansible_it_fabric }}"
        deploy: false
        state: merged
        config:
          - name: add_ip_prefix_list              # This must be a valid template name
            create_additional_policy: false  # Do not create a policy if it already exists
            priority: 101
            policy_vars:
              PREFIX_LIST_NAME: ansible-prefix-list
              ACL_NAME: 10.122.84.0/24
              SEQ_NUM: 20
              PREFIX_LIST_ACTION: permit

    
          - switch:
              - ip: "{{ ansible_switch1 }}"

      vars:
          ansible_command_timeout: 1000
          ansible_connect_timeout: 1000
```

### Debug Output

https://gist.github.com/micruzz82/15a0992561c4584f8b962abe2305b32a

### Expected Behavior

[...]

PR #67, opened 2021-04-30
Policy module fix for modify with template name

mmudigon:
This PR fixes a couple of issues in dcnm_policy module. (https://github.com/CiscoDevNet/ansible-dcnm/issues/60)

1. When a policy is created (state = merged) using a template name that is already used by another policy, the new policy is not created. Because the existing code allows modifications only with policy-ids. This PR updates the code to fix this which will now allow the new policy to be created instead of treating it like a modify operation.
2. When a policy is deleted, all the data from DCNM server is deleted appropriately. But the configuration related to the policy is not removed from the switch. This PR fixes an issue with the mark-delete path which has changed in 11.5. With this fix the policy configuration is removed from the switch.


tests/integration/targets/dcnm_policy/tests/dcnm/dcnm_policy_delete.yaml
```diff
@@ -106,28 +106,7 @@
         that:
           - 'item["RETURN_CODE"] == 200'
           - 'item["MESSAGE"] == "OK"'
-      when: (my_idx < (result["diff"][0]["deleted"] | length))
```
mikewiebe, line 109: Can you add test cases to explicitly cover the following:

* Create several policies with a template1 (no variables)
* Add new policies using template1 where `description` and `priority` are different then existing policies
* Force creation of new policies with the force flag set to true
* Verify policies are not created when the force flag is false and policies match already existing policies
* Repeat this for template2 (with variables)

mikewiebe: Some of these might already be covered but want to make sure we have coverage for all of them

mmudigon: Mike these are already covered as part of some of the other cases. Anyway I added some more cases to explicitly cover the new cases. Pushed the changes again.

mikewiebe:
@mmudigon This is looking much better but I found a couple of issues:

1. I can't figure out how to delete a single policy.  When I specify a policy name it still deletes all of the policies associated with the template.
    * IMO we should be able to delete a single policy or targeted set of policies when the name is the name of the policy id
    * Also, since the default behavior seems to be to delete all policies, we should be able to omit the `config:` keyword
2. When the `deploy` flag is set to true I see the `changed` flag being set even if no new policies are being created

Issue #66, opened 2021-04-30
remote as in vrf lite is hard cored to 65535


dsx1123:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Ansible Version and collection version

```
ansible 2.10.5
  config file = /root/.ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /root/virtualenv/ansible/lib/python3.7/site-packages/ansible
  executable location = /root/virtualenv/ansible/bin/ansible
  python version = 3.7.3 (default, Apr  3 2019, 05:39:12) [GCC 8.3.
```
### DCNM version

* V 11.5.1

### Affected module(s)

* dcnm_vrf

### Ansible Playbook

```yaml
# Copy-paste your anisble playbook here 

```
same play as issue #65 
### Debug Output

same output as #65 
### Expected Behavior

the remote as should be set to what is configured in external fabric, in this case, it is `200`

### Actual Behavior

After push the configuration to border leaf, this configuration shows in border:
```
 neighbor 10.31.0.2
      remote-as 65535
      address-family ipv4 unicast
        send-community
        send-community extended
        route-map extcon-rmap-filter out
```

### References
looks like we hard coded the as number to 65535:
```python
vrflite_con['VRF_LITE_CONN'][0]['NEIGHBOR_ASN']='65535'
```

PR #68, opened 2021-05-04
Fix for neighbor_asn in vrflite

praveenramoorthy:
This PR fixes the issue in setting neighbor_asn for vrflite in #66


mikewiebe (commented):
Looks good.  Can we add a test to cover this change if we don't already have one?

PR #81, opened 2021-07-13
Dcnm network merge fix

praveenramoorthy:
This PR fixes the issue in merge operation of Network module

### DIFF
--- a/plugins/modules/dcnm_network.py
+++ b/plugins/modules/dcnm_network.py
@@ -23,7 +23,7 @@ from ansible_collections.cisco.dcnm.plugins.module_utils.network.dcnm.dcnm impor
 from ansible.module_utils.connection import Connection
 from ansible.module_utils.basic import AnsibleModule
 
-__author__ = "Chris Van Heuveln, Shrishail Kariyappanavar, Karthik Babu Harichandra Babu"
+__author__ = "Chris Van Heuveln, Shrishail Kariyappanavar, Karthik Babu Harichandra Babu, Praveen Ramoorthy"
 
 DOCUMENTATION = '''
 ---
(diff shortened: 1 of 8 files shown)


plugins/modules/dcnm_network.py
```diff
@@ -1975,6 +1975,133 @@ def failure(self, resp):
 
         self.module.fail_json(msg=res)
 
+    def dcnm_update_network_information(self, want, have, cfg):
+
+        if cfg.get("vrf_name", None) is None:
```
mmudigon, line 1980: We need not compare mandatory objects here because they have to be included. Just check

tests/integration/targets/dcnm_network/tests/dcnm/merged.yaml
```diff
@@ -645,6 +645,121 @@
     fabric: "{{ ansible_it_fabric }}"
     state: deleted
 
+- name: MERGED - Create L3 networks along with all dhcp, arp options without attach
+  cisco.dcnm.dcnm_network:
```
mmudigon, line 649: Looks like this is a new test case. Did you check if the existing test cases for merge work properly?

plugins/modules/dcnm_network.py
praveenramoorthy: Network name is the only mandatory parameter.

tests/integration/targets/dcnm_network/tests/dcnm/merged.yaml
praveenramoorthy: Yes all the exiting cases are fine. This specific scenario where we create network first and attach switched subsequently in a different operation was not cover before for merge, so added a case for that.

plugins/modules/dcnm_network.py
```diff
@@ -1975,6 +1975,133 @@ def failure(self, resp):
...
+            json_to_dict_want["tag"] = json_to_dict_have["tag"]
+            if json_to_dict_want["tag"] != "":
+                json_to_dict_want["tag"] = int(json_to_dict_want["tag"])
+
+        if cfg.get("gw_ip_subnet", None) is None:
+            json_to_dict_want["gatewayIpAddress"] = json_to_dict_have["gatewayIpAddress"]
+
+        if cfg.get("is_l2only", None) is None:
+            json_to_dict_want["isLayer2Only"] = json_to_dict_have["isLayer2Only"]
+            if json_to_dict_want["isLayer2Only"] == "true":
```
mikewiebe, line 2010: Just to be safe you might want to use the to_lower() function.  That way of the string were to ever change to `"True"` for instance it would not break.

```python
if json_to_dict_want["isLayer2Only"].to_lower() == "true":
```

PR #82, opened 2021-07-14
Vrf name input change for L2 networks

praveenramoorthy:
This PR adds the capability to specify vrf_name as "" for L2 networks. For L2 networks, user can either specify vrf_name as "" or they can avoid specifying vrf_name at all

### DIFF
--- a/plugins/module_utils/network/dcnm/dcnm.py
+++ b/plugins/module_utils/network/dcnm/dcnm.py
@@ -74,6 +74,8 @@ def validate_list_of_dicts(param_list, spec, module=None):
                     if spec[param].get('length_max'):
                         if 1 <= len(item) <= spec[param].get('length_max'):
                             pass
+                        elif param == "vrf_name" and (len(item) <= spec[param].get('length_max')):
+                            pass
                         else:
                             invalid_params.append('{}:{} : The string exceeds the allowed '
                                                   'range of max {} char'.format(param, item,
--- a/plugins/modules/dcnm_network.py
+++ b/plugins/modules/dcnm_network.py
@@ -65,7 +65,8 @@ options:
       vrf_name:
         description: 'Name of the VRF to which the network belongs to'
         type: str
-        note: This field is required for L3 Networks
+        note: This field is required for L3 Networks. VRF name should not be specified
+              or may be specified as "" for L2 networks
       net_id:
         description: 'ID of the network being managed'
         type: int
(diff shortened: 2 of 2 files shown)


mikewiebe (approved):

dsx1123:
@praveenramoorthy  the empty vrf_name is not accepted by DCNM, from the backend, I believe we should replace it with NA. otherwise, it will give the error:

mikewiebe:
> @praveenramoorthy the empty vrf_name is not accepted by DCNM, from the backend, I believe we should replace it with NA. otherwise, it will give the error:
> 
> ```
> failed: [10.211.105.105] (item={'network_name': 'N0010', 'vrf_name': '', 'vlan_id': 2301, 'gw_ip_subnet': '', 'leaf': '10.211.105.123', 'ports': 'Ethernet1/1', 'l3mtu': 9216, 'network_id': 35655, 'is_l2only': 'yes', 'arp_suppress': 'no'}) => {"ansible_loop_var": "item", "changed": false, "item": {"arp_suppress": "no", "gw_ip_subnet": "", "is_l2only": "yes", "l3mtu": 9216, "leaf": "10.211.105.123", "network_id": 35655, "network_name": "N0010", "ports": "Ethernet1/1", "vlan_id": 2301, "vrf_name": ""}, "msg": "VRF: is missing in fabric: DN-LAB-FABRIC"}
> ```

Thanks for following up @dsx1123.  This is effectively what @praveenramoorthy is doing in the new diffs

https://github.com/CiscoDevNet/ansible-dcnm/pull/82/files#diff-81913290492fc2a4c62ca8ade79047040d6afc496e3dbc92761e14a1ed6f472cR815

dsx1123:
@mikewiebe I am using the latest commit to test it, but still failed.

mikewiebe:
> @mikewiebe I am using the latest commit to test it, but still failed.

Got it.  Sorry for the confusion.

dsx1123:
@mikewiebe in the playbook, I set is_l2only to `"yes"` by mistake,  where it should be `yes`. somehow the ansible accept the `"yes"` as boolean type but didn't convert it to True in python, if I  use `is_l2only: yes` or `is_l2only: True`, it does work. It doesn't looks like the modules' issue but more like on ansibe side.

praveenramoorthy:
@dsx1123 It works fine for "yes" in my python env. I think this issue is specific to env(may be because of older versions of libraries)

dsx1123:
> @dsx1123 It works fine for "yes" in my python env. I think this issue is specific to env(may be because of older versions of libraries)

it could be, here is my envrionement, just for your reference:

Issue #101, opened 2021-10-14
query parameter for vrf and network does not provide information


qsnyder:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Ansible Version and collection version

ansible 2.10.10
collection version 1.2.1

### DCNM version

* V 11.5.1

### Affected module(s)

* dcnm_vrf
* dcnm_module

### Ansible Playbook

```yaml
---

- hosts: dcnm_controllers
  gather_facts: false

  vars:
    fabric: DevNet_Fabric

  tasks:
    - name: Query VRF deployed from Ansible
      cisco.dcnm.dcnm_vrf:
        fabric: "{{ fabric }}"
        state: query
        config:
        - vrf_name: Ansible_VRF
          vrf_template: Default_VRF_Universal
          vrf_extension_template: Default_VRF_Extension_Universal
```

### Expected Behavior

I'd expect something querying the specific URI for the VRF or network and returning the configured parameters of such.  Similar to this returned from `/top-down/fabrics/{fabric}/vrfs/{vrf-name}`

```
{
  "fabric": "DevNet_Fabric",
  "vrfName": "Ansible_VRF",
  "vrfTemplate": "Default_VRF_Universal",
  "vrfExtensionTemplate": "Default_VRF_Extension_Universal",
  "vrfTemplateConfig": "{\"advertiseDefaultRouteFlag\":\"true\",\"vrfVlanId\":\"\",\"isRPExternal\":\"false\",\"bgpPassword\":\"\",\"vrfSegmentId\":\"50000\",\"multicastGroup\":\"\",\"configureStaticDefaultRouteFlag\":\"true\",\"advertiseHostRouteFlag\":\"false\",\"trmEnabled\":\"false\",\"trmBGWMSiteEnabled\":\"false\",\"asn\":\"65001\",\"nveId\":\"1\",\"vrfName\":\"Ansible_VRF\"}",
  "tenantName": null,
  "vrfId": 50000,
  "serviceVrfTemplate": null,
  "source": null
}
```

### Actual Behavior

Return looks similar to this

```
ok: [dcnm] => {
[...]

PR #111, opened 2021-11-17
Vrf name and network name for query with config

praveenramoorthy:
Changes to support Vrf name alone to query vrf with config. 
Changes to Network name alone to query network with config.
Fixes the issue in #101.

### DIFF
--- a/plugins/modules/dcnm_vrf.py
+++ b/plugins/modules/dcnm_vrf.py
@@ -1294,17 +1294,6 @@ class DcnmVrf:
 
                     if (want_c['vrfName'] == vrf['vrfName']):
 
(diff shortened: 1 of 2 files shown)


plugins/modules/dcnm_vrf.py
```diff
@@ -335,16 +335,7 @@
     state: query
     config:
     - vrf_name: ansible-vrf-r1
-      vrf_id: 9008011
```
mikewiebe, line 338: Just double checking.  I know that this PR enabled a query using just vrf or network but does it also support other parameters as additional filters?

Are we still testing other filters?

PR #116, opened 2021-11-29
policy module fix

mmudigon:
Updated dcnm_policy module to perform retries to deploy policies.

### DIFF
--- a/plugins/modules/dcnm_policy.py
+++ b/plugins/modules/dcnm_policy.py
@@ -339,6 +339,7 @@ EXAMPLES = """
 import json
 import re
 import copy
+import time
 
 from ansible.module_utils.basic import AnsibleModule
 from ansible_collections.cisco.dcnm.plugins.module_utils.network.dcnm.dcnm import (
@@ -360,7 +361,8 @@ class DcnmPolicy:
               "POLICY_MARK_DELETE": "/rest/control/policies/{}/mark-delete",
               "POLICY_DEPLOY": "/rest/control/policies/deploy",
               "POLICY_CFG_DEPLOY": "/rest/control/fabrics/{}/config-deploy/",
-              "POLICY_WITH_POLICY": "/rest/control/policies/{}"
+              "POLICY_WITH_POLICY_ID": "/rest/control/policies/{}",
+              "CONFIG_PREVIEW": "/rest/control/fabrics/{}/config-preview?forceShowRun=false&showBrief=true"
             },
         12: {
               "POLICY_WITH_ID": "/appcenter/cisco/ndfc/v1/lan-fabric/rest/control/policies/{}",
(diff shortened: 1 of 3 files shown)


plugins/modules/dcnm_policy.py
```diff
@@ -927,7 +930,7 @@ def dcnm_policy_delete_policy(self, policy, mark_del):
             command = "PUT"
         else:
             #path = "/rest/control/policies/" + policy
-            path = self.paths["POLICY_WITH_POLICY"].format(policy)
+            path = self.paths["POLICY_WITH_POLICY_ID"].format(policy)
```
mikewiebe, line 933: What is the reason for changing from `POLICY_WITH_POLICY` to `POLICY_WITH_POLICY_ID`?

plugins/modules/dcnm_policy.py
```diff
@@ -994,18 +997,40 @@ def dcnm_policy_send_message_to_dcnm(self):
...
+
+                retries += 1
+                resp = self.dcnm_policy_save_and_deploy(snos)
+                if (
+                    resp
+                    and (resp["RETURN_CODE"] != 200)
+                ):
+                    self.module.fail_json(msg=resp)
+
+                path = self.paths["CONFIG_PREVIEW"].format(self.fabric)
```
mikewiebe, line 1012: What does `CONFIG_PREVIEW` indicate?  Can you add some comments here?

tests/unit/modules/dcnm/fixtures/dcnm_policy_payloads.json
```diff
@@ -1537,5 +1537,30 @@
       "METHOD": "DELETE",
       "REQUEST_PATH": "https://10.64.78.151:443/rest/control/policies/POLICY-105105",
       "RETURN_CODE": 200
+    },
+
+    "config_preview": {
+      "RETURN_CODE": 200,
+      "METHOD": "GET",
+      "REQUEST_PATH": "https://10.195.225.193:443/appcenter/cisco/ndfc/api/v1/lan-fabric/rest/control/fabrics/mmudigon/config-preview?forceShowRun=false&showBrief=true",
```
mikewiebe, line 1545: Better to use a private IP for the unit test.

plugins/modules/dcnm_policy.py
mmudigon: It is actually getting Policy information using POLICY-ID. So changed it to reflect it correctly

plugins/modules/dcnm_policy.py
mmudigon: It gets the current status of the confguration on the DCNM. If it is not "In-Sync" then there re more commands to be deployed. So we keep retrying

mmudigon: Added comments for this

tests/unit/modules/dcnm/fixtures/dcnm_policy_payloads.json
mmudigon: Done

PR #124, opened 2021-12-06
Add dhcp_loopback_id parameter support

mikewiebe:
This is a cherry-pick of the recent feature support I added and published as a hot-fix to Ansible Galaxy.  We don't have this feature in develop so this is essentially a double commit.

Original commit here: https://github.com/CiscoDevNet/ansible-dcnm/commit/a7de8a969bb00bac2af2843f6e6add87997e51fc

There are also a couple of formatting changes that my IDE picked up and changed.

### DIFF
--- a/plugins/module_utils/network/dcnm/dcnm.py
+++ b/plugins/module_utils/network/dcnm/dcnm.py
@@ -83,8 +83,11 @@ def validate_list_of_dicts(param_list, spec, module=None):
                                                                                 spec[param].get('length_max')))
                 elif type == 'int':
                     item = v.check_type_int(item)
+                    min_value = 1
+                    if spec[param].get('range_min') is not None:
+                        min_value = spec[param].get('range_min')
                     if spec[param].get('range_max'):
-                        if 1 <= item <= spec[param].get('range_max'):
+                        if min_value <= item <= spec[param].get('range_max'):
                             pass
                         else:
                             invalid_params.append('{}:{} : The item exceeds the allowed '
--- a/docs/cisco.dcnm.dcnm_network_module.rst
+++ b/docs/cisco.dcnm.dcnm_network_module.rst
@@ -170,6 +170,22 @@ Parameters
                     <td class="elbow-placeholder"></td>
                 <td colspan="2">
                     <div class="ansibleOptionAnchor" id="parameter-"></div>
+                    <b>dhcp_loopback_id</b>
+                    <a class="ansibleOptionLink" href="#parameter-" title="Permalink to this option"></a>
+                    <div style="font-size: small">
+                        <span style="color: purple">integer</span>
+                    </div>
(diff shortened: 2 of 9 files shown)


praveenramoorthy (approved):

plugins/module_utils/network/dcnm/dcnm.py
```diff
@@ -83,8 +83,11 @@ def validate_list_of_dicts(param_list, spec, module=None):
                                                                                 spec[param].get('length_max')))
                 elif type == 'int':
                     item = v.check_type_int(item)
+                    min_value = 1
+                    if spec[param].get('range_min') is not None:
+                        min_value = spec[param].get('range_min')
                     if spec[param].get('range_max'):
-                        if 1 <= item <= spec[param].get('range_max'):
+                        if min_value <= item <= spec[param].get('range_max'):
```
mikewiebe, line 90: @mmudigon This change is causing a unit testcase for the `dcnm_interface` module to fail.  I did some investigation and discovered that it's because the `range_min` for the `ipv6_mask_len` parameter is `64` but one of the unit test cases uses a length of `48`. 

Is `48` a valid mask length?  

We need to either:

* Change the `range_min` for this parameter in the dcnm_interface module to something lower or...
* Change the test case to use a value above 48.

mikewiebe: If I update the fixture file to this the test passes:

```diff
diff --git a/tests/unit/modules/dcnm/fixtures/dcnm_intf_subint_configs.json b/tests/unit/modules/dcnm/fixtures/dcnm_intf_subint_configs.json
index 21c84fd..4f9cb91 100644
--- a/tests/unit/modules/dcnm/fixtures/dcnm_intf_subint_configs.json
+++ b/tests/unit/modules/dcnm/fixtures/dcnm_intf_subint_configs.json
@@ -172,7 +172,7 @@
                                "ipv6_addr": "fe80::02",
                                "int_vrf": "default",
                                "ipv4_mask_len": 16,
-                               "ipv6_mask_len": 48,
+                               "ipv6_mask_len": 68,
                                "fabric": "test_fabric",
                                "sno": "SAL1819SAN8",
                                "vlan": 201,
```

mikewiebe: I made the update in the PR to fix the failing test but we should discuss what the correct fix should be:

https://github.com/CiscoDevNet/ansible-dcnm/pull/124/commits/d53b1f9339af3af26d7ef33c85887ce43f957e78

mmudigon: Though 48 is a valid IPv6 mask length, the minimum value suppported by DCNM is 64. So your fix is ok.

mmudigon (approved):

PR #129, opened 2022-01-27
Changes for latest NDFC login

praveenramoorthy:
Changes to support the latest NDFC login

### DIFF
--- a/plugins/httpapi/dcnm.py
+++ b/plugins/httpapi/dcnm.py
@@ -130,19 +130,16 @@ class HttpApi(HttpApiBase):
         method = 'POST'
         path = {'dcnm': '/rest/logon', 'ndfc': '/login'}
         login12Func = [self._login_latestv2, self._login_latestv1]
-        iter = 0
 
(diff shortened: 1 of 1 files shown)


plugins/httpapi/dcnm.py
```diff
@@ -108,13 +129,20 @@ def login(self, username, password):
...
         path = {'dcnm': '/rest/logon', 'ndfc': '/login'}
+        login12Func = [self._login_latestv2, self._login_latestv1]
+        iter = 0
 
         # Attempt to login to DCNM version 11
         self._login_old(username, password, method, path['dcnm'])
 
         # If login attempt failed then try NDFC version 12
         if not self.login_succeeded:
-            self._login_latest(username, password, method, path['ndfc'])
```
mikewiebe, line 117: I believe you could simplify this a bit and avoid needing an extra variable `iter`.

```python
for func in login12Func:
    func(username, password, method, path['ndfc'])
    if self.login_succeeded:
        break
```

mikewiebe (approved):

PR #135, opened 2022-02-18
Add Python template type support

rost-d:
Adding capability to create python type templates via new config option


mikewiebe:
@mmudigon GitHub action tests are now passing, can you please review?

mmudigon (approved):

mikewiebe:
@rost-d This generally looks ok but just wanted to make sure that the changes is idempotent.  We have integration tests for cli type but since you are adding a new type just want to make sure.  Have you tested that?

Issue #139, opened 2022-02-21
Policy Description filed is set to blank when modify policy is used


risarwarcisco:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Ansible Version and collection version

ansible 2.9.27
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/risarwar/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.6/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.6.8 (default, Sep 10 2021, 09:13:53) [GCC 8.5.0 20210514 (Red Hat 8.5.0-3)]

### DCNM version

* V 11.5(3a)

### Affected module(s)

* dcnm_policy

### Ansible Playbook

```yaml
 - name: Configure Mulitple Policies 
        cisco.dcnm.dcnm_policy:
            fabric: "{{ fabric }}"
            deploy: false
            state: merged
            config:
              - switch:
                  - ip: "{{ Switch_name }}"
                    policies:
                        - name: "{{item.policy_id}}"
                          create_additional_policy: false
                          priority: "{{item.priority | default(500)}}"
                          description: "{{item.description}}"
                          policy_vars: "{{ item.Values | default([])}}"
        register: policy_update_data
        loop: "{{ Policies }}"
        ignore_errors: yes
```

### Expected Behavior

I expect this play to update the Policy with specified policy ID and not change fields that have not been changed.

### Actual Behavior

The description field is set to null (when it was previously set upon creation of the policy). Even when the description field is not specified, same result is produced.

mikewiebe:
@risarwarcisco Thanks for raising the issue.  We are investigating.

PR #142, opened 2022-03-16
Fix for issue 139 - Description is null during update policy

mmudigon:
This PR fixes issue https://github.com/CiscoDevNet/ansible-dcnm/issues/139 

DESCRIPTION:

The issue is when an existing policy is updated, the description field is becoming empty. 

CAUSE:

This is because, when a modify payload is prepared, the description field is left out and is defaulted to empty.

FIX:

Added code to copy the description appropriately during update


plugins/modules/dcnm_policy.py
```diff
@@ -732,9 +732,20 @@ def dcnm_policy_get_diff_merge(self):
...
+                # is requested to be created. Check the 'create_additional_policy' flag and create it if it is
                 # set to True
                 if policy["create_additional_policy"] is True:
+
+                    # Check if policy is being created using policy ID. In such a case since we are trying to create
+                    # and additional policy here, clear the "id" and "policyId" fields. Otherwise CREATE will fail
+                    # complaining the policy being created is not unique since a policy with the same policy number
+                    # already exists
+
+                    if "POLICY-" in policy.get("policyId", ''):
```
mikewiebe, line 744: Is this part of the fix for description or is this a different fix you found while testing?

plugins/modules/dcnm_policy.py
```diff
@@ -732,9 +732,20 @@ def dcnm_policy_get_diff_merge(self):
...
+                        policy.pop("id")
+                        policy.pop("policyId")
+                        policy["policy_id_given"] = False
+
                     self.changed_dict[0]["merged"].append(policy)
```
mikewiebe, line 749: Can you add unit tests to cover this fix?

mikewiebe (changes requested):

plugins/modules/dcnm_policy.py
mmudigon: This was a corner case and users should not be using policy-id to create a new policy. But in case they do we will cover that with this fix

plugins/modules/dcnm_policy.py
mmudigon: Modified existing It and UT to check for this scenario

PR #147, opened 2022-04-07
Resource Manager module initial checkin

mmudigon:
This checkin includes the following:

    - Resource Manager code to Create, Delete and Query resources
    - Documentation updates for resource manager 
    - Documentation updates for other modules
    - IT and UT infra along with the test cases


plugins/modules/dcnm_res_manager.py
```diff
@@ -0,0 +1,1254 @@
+#!/usr/bin/python
+#
+# Copyright (c) 2020-2022 Cisco and/or its affiliates.
```
mikewiebe, line 3: Since this is a new module we can set the copyright to just `2022` and extend it later if modified.

plugins/modules/dcnm_res_manager.py
```diff
@@ -0,0 +1,1254 @@
...
+        - IP address or DNS name of the management interface of the switch to which the allocated resource is assigned to.
+        type: list
+        required: false
+"""
+
+EXAMPLES = """
+# Entity name format
+# ==================
+#
+# The formt of the entity name depends on the scope_type of the resource being allocated.
```
mikewiebe, line 103: `formt/format`

plugins/modules/dcnm_res_manager.py
```diff
@@ -0,0 +1,1254 @@
...
+
+# Merged:
+#   Resources defined in the playbook will be merged into the target fabric.
+#     - If the Resources does not exist it will be added.
+#     - If the Resources exists but properties managed by the playbook are different
+#       they will be updated if possible.
+#     - Resources that are not specified in the playbook will be untouched.
+#
+# Deleted:
+#   Resources defined in the playbook will be deleted.
```
mikewiebe, line 125: Do we get an error message informing the user if we don't specify a resource with state deleted?   Just want to make sure since we support this in some module but not this one.

plugins/modules/dcnm_res_manager.py
```diff
@@ -0,0 +1,1254 @@
...
+
+# DELETING RESOURCES
+# ==================
+
+- name: Delete Resources
+  cisco.dcnm.dcnm_res_manager:
+    state: deleted                              # choose form [merged, deleted, query]
+    fabric: test_fabric
+    config:
+      - entity_name: "l3_vni_fabric"            # A unique name to identify the resource
```
mikewiebe, line 182: Can we delete the resource using just the `entity_name`?

plugins/modules/dcnm_res_manager.py
```diff
@@ -0,0 +1,1254 @@
...
+        description:
+          - Name of the resource pool from which the resource is allocated
+        type: str
+        required: true
+      scope_type:
+        description:
+          - Socpe of resource allocation
+        type: str
+        required: true
+        choices:
```
mikewiebe, line 77: Maybe I missed this in the code but are we enforcing these choices in the module?

plugins/modules/dcnm_res_manager.py
```diff
@@ -0,0 +1,1254 @@
...
+          - The format of this parameter depends on the scope_type. The details are provided in
+          - the EXAMPLES section
+        type: str
+        required: true
+      pool_type:
+        description:
+          - Type of resource pool
+        type: str
+        required: true
+        choices:
```
mikewiebe, line 63: Maybe I missed this in the code but are we enforcing these choices in the module with the argument spec?

tests/integration/targets/dcnm_res_manager/tests/dcnm/dcnm_res_manager_invalid_params.yaml
```diff
@@ -0,0 +1,130 @@
...
+      cisco.dcnm.dcnm_res_manager:
+        state: merged                                # choose form [merged, deleted, query]
+        fabric: "{{ ansible_it_fabric }}" 
+        config:
+          - entity_name: "l3_vni_fabric"             # A unique name to identify the resource
+            pool_type: "IDLE"                          # choose from ['ID', 'IP', 'SUBNET']
+            pool_name: "L3_VNI"                  # Based on the 'poolType', select appropriate name
+            scope_type: "fabric"                     # choose from ['fabric', 'device', device_interface', 'device_pair', 'link']             
+            resource: "101"                          # The value of the resource being created
+      ignore_errors: yes
```
mikewiebe, line 24: I see you are using the `ignore_errors` flag since these are invalid cases but should we also be storing the result and checking the error message or at least that an error was generated?

tests/integration/targets/dcnm_res_manager/tests/dcnm/dcnm_res_manager_merge.yaml
```diff
@@ -0,0 +1,257 @@
...
+
+- name: Remove local log file
+  local_action: command rm -f res_mgr.log
+
+- name: Delete Resources
+  cisco.dcnm.dcnm_res_manager: &rm_delete
+    state: deleted                                 # choose form [merged, deleted, query]
+    fabric: "{{ ansible_it_fabric }}" 
+    config:
+      - entity_name: "l3_vni_fabric"               # A unique name to identify the resource
```
mikewiebe, line 13: Please add a testcase to delete with just the `entity_name` if this is supported.

tests/integration/targets/dcnm_res_manager/tests/dcnm/dcnm_res_manager_query.yaml
```diff
@@ -0,0 +1,302 @@
...
+            switch:                                  # provide the switch information to which the given resource is attached
+              - "{{ ansible_switch1 }}"              
+          - pool_name: "SUBNET"                      # Based on the 'poolType', select appropriate name
+            switch:                                  # provide the switch information to which the given resource is attached
+              - "{{ ansible_switch1 }}"
+      register: result
+
+    - assert:
+        that:
+          - '(result["response"] | length) != 0'
```
mikewiebe, line 257: Should we be checking for specific information in the query result?

plugins/modules/dcnm_res_manager.py
mmudigon: Done

plugins/modules/dcnm_res_manager.py
mmudigon: Done

plugins/modules/dcnm_res_manager.py
mmudigon: As discussed, all parameters are mandatory for state DELETED. So if a parameter is not included it will error out.

plugins/modules/dcnm_res_manager.py
mmudigon: No. All parameters are mandatory

plugins/modules/dcnm_res_manager.py
mmudigon: Yes, We have a routine which checks for allowed combinations

plugins/modules/dcnm_res_manager.py
mmudigon: Yes, We have a routine which checks for allowed combinations

tests/integration/targets/dcnm_res_manager/tests/dcnm/dcnm_res_manager_invalid_params.yaml
mmudigon: Will add the checks

tests/integration/targets/dcnm_res_manager/tests/dcnm/dcnm_res_manager_merge.yaml
mmudigon: Not supported. all parameters are mandatory

tests/integration/targets/dcnm_res_manager/tests/dcnm/dcnm_res_manager_query.yaml
mmudigon: I can check for the requested resources. but in case it is a query all, we will not know what all resources will be returned. So there is no way for us to check for specific names. But for specific test cases requesting for specific resources, we can check.

README.md
```diff
@@ -36,6 +36,7 @@ Name | Description
 [cisco.dcnm.dcnm_inventory](https://github.com/CiscoDevNet/ansible-dcnm/blob/main/docs/cisco.dcnm.dcnm_inventory_module.rst)|Add and remove Switches from a DCNM managed VXLAN fabric.
 [cisco.dcnm.dcnm_network](https://github.com/CiscoDevNet/ansible-dcnm/blob/main/docs/cisco.dcnm.dcnm_network_module.rst)|Add and remove Networks from a DCNM managed VXLAN fabric.
 [cisco.dcnm.dcnm_policy](https://github.com/CiscoDevNet/ansible-dcnm/blob/main/docs/cisco.dcnm.dcnm_policy_module.rst)|DCNM Ansible Module for managing policies.
+[cisco.dcnm.dcnm_res_manager](https://github.com/CiscoDevNet/ansible-dcnm/blob/main/docs/cisco.dcnm.dcnm_res_manager_module.rst)|DCNM ansible module for managing resources.
```
mikewiebe, line 39: @mmudigon I would prefer we expand `res` to read `resource`.  It's not that much longer and much less ambiguous

`dcnm_resource_manager`

PR #156, opened 2022-07-29
Switch level deploy instead of fabric level deploy for policy module

mmudigon:
In the DCNM module there is no option to push individual policy changes during DELETE operations. In the current implementation, a Fabric level config deploy is executed after DELETE of Policies. In large fabrics this may take considerable time (because of fabric level deploy) even if the DELETE operation is restricted to a few switches. 
To solve this problem, once a Policy is deleted, deploy operation is executed only for the specific switches which are impacted by the DELETE. This way the operation will be much quicker.

NOTE: Even if a single policy is deleted, a switch level deploy will deploy all the pending configuration with respect to the switch.

### DIFF
--- a/plugins/modules/dcnm_policy.py
+++ b/plugins/modules/dcnm_policy.py
@@ -414,6 +414,9 @@ class DcnmPolicy:
         self.inventory_data = get_fabric_inventory_details(
             self.module, self.fabric
         )
+
+        # Get all switches which are managable. Will be required to check deploy status.
+        self.managable = [self.inventory_data[key]["serialNumber"] for key in self.inventory_data.keys() if self.inventory_data[key]["managable"]]
         self.ip_sn, self.hn_sn = get_ip_sn_dict(self.inventory_data)
 
         self.result = dict(changed=False, diff=[], response=[])
@@ -989,7 +992,9 @@ class DcnmPolicy:
 
             # Get all serial numbers. We will require this to do save and deploy
             if policy["serialNumber"] not in snos:
-                snos.append(policy["serialNumber"])
+                # Add the serial number only if the switch is managable.
+                if policy["serialNumber"] in self.managable:
+                    snos.append(policy["serialNumber"])
 
             # First Mark the policy as deleted. Then deploy the same to remove the configuration
             # from the switch. Then we can finally delete the policies from the DCNM server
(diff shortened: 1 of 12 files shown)


plugins/modules/dcnm_policy.py
```diff
@@ -365,7 +367,7 @@ class DcnmPolicy:
             "POLICY_DEPLOY": "/rest/control/policies/deploy",
             "POLICY_CFG_DEPLOY": "/rest/control/fabrics/{}/config-deploy/",
             "POLICY_WITH_POLICY_ID": "/rest/control/policies/{}",
-            "CONFIG_PREVIEW": "/rest/control/fabrics/{}/config-preview?forceShowRun=false&showBrief=true",
+            "CONFIG_PREVIEW": "/rest/control/fabrics/{}/config-preview/",
```
mikewiebe, line 370: What is the functional difference between these two?

plugins/modules/dcnm_policy.py
```diff
@@ -438,7 +442,9 @@ def dcnm_policy_validate_input(self):
 
         policy_spec = dict(
             name=dict(required=True, type="str"),
-            create_additional_policy=dict(required=False, type="bool", default=True),
+            create_additional_policy=dict(
```
mikewiebe, line 445: Looks like there are quite a few formatting only changes in this diff set.  I assume this was from running `ansible-test` ?

plugins/modules/dcnm_policy.py
```diff
@@ -1070,9 +1104,9 @@ def dcnm_policy_send_message_to_dcnm(self):
                 and (resp.get("DATA", None) is not None)
             ):
                 if resp["DATA"].get("successList", None) is not None:
-                    if "is created successfully" in resp["DATA"]["successList"][0].get(
-                        "message"
-                    ):
+                    if "is created successfully" in resp["DATA"][
```
mikewiebe, line 1107: Would it be better to check the return code here instead of the message?  Just worried that the string message could change in the future and this check would fail

plugins/modules/dcnm_policy.py
mmudigon: The removed one was for getting all preview of all switches in the fabric which may not be scalable. So we now moved to the new one, where we specify switch serial numbers which are of interest. So the string "forceShowRun=false&showBrief=true" should come after the switch serial numbers. So we moved adding this to routine where we add switch info to the API

plugins/modules/dcnm_policy.py
mmudigon: yes

plugins/modules/dcnm_policy.py
mmudigon: this is the only string available and it is same between versions. Otherwise we may have a RETURN CODE 200, but there may be a case of failure. Added this specifically to check for that. Also this was already there in the released code.

mikewiebe (approved):

PR #158, opened 2022-08-18
DCNM Links module initial checkin

mmudigon:
This PR commits the code changes to support DCNM Links Module operations. It includes
    - Module backend code
    - Unit test code
    - Unit Test Cases
    - IT playbooks


plugins/modules/dcnm_links.py
```diff
@@ -0,0 +1,3326 @@
...
+__metaclass__ = type
+__author__ = "Mallik Mudigonda"
+
+DOCUMENTATION = """
+---
+module: dcnm_links
+short_description: DCNM ansible module for managing Links.
+version_added: "2.1.0"
+description:
+    - DCNM ansible module for creating, modifying, deleting and querying Links
```
mikewiebe, line 28: Let's check with Shangxin but I think we need to start using NDFC instead of DCNM as the module names and documentation references.

plugins/modules/dcnm_links.py
```diff
@@ -0,0 +1,3326 @@
...
+    module = AnsibleModule(
+        argument_spec=element_spec, supports_check_mode=True
+    )
+
+    dcnm_links = DcnmLinks(module)
+
+    state = module.params["state"]
+
+    if [] is dcnm_links.config:
+        if state == "merged":
```
mikewiebe, line 3274: This should also be mandatory for `replaced` right?

plugins/modules/dcnm_links.py
mmudigon: Sure. Let me know. I can change it

plugins/modules/dcnm_links.py
```diff
@@ -0,0 +1,3326 @@
...
+            required: true
+          peer2_ipv4_address:
+            description:
+              - IPV4 address of the destination interface.
+              - This parameter is optional if the underlying fabric is ipv6 enabled.
+              - This parameter is required only if template is 'int_intra_fabric_num_link' or
+                'ios_xe_int_intra_fabric_num_link' or 'int_intra_vpc_peer_keep_alive_link'.
+            type: str
+            required: true
+          peer1_ipv6_address:
```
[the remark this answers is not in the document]
mmudigon: Even in GUI this is optional if the fabric is not IPV6 enabled.  Since this is optional param, user can still configure the IPv6 addresses if that is the only address available

plugins/modules/dcnm_links.py
mmudigon: yes. you are right. Will update

plugins/modules/dcnm_links.py
mikewiebe: Before we make any changes though we need to check if we can alias based on the old name.

https://ansible-docs.readthedocs.io/zh/stable-2.0/rst/developing_modules.html#deprecating-and-making-module-aliases

Issue #168, opened 2022-10-03
handle not found fabric gracefully on dcnm_inventory


dsx1123:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Ansible Version and collection version
```
ansible [core 2.11.10]
  config file = /root/.ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /root/virtualenv/ansible/lib/python3.9/site-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /root/virtualenv/ansible/bin/ansible
  python version = 3.9.2 (default, Feb 28 2021, 17:03:44) [GCC 10.2.1 20210110]
  jinja version = 3.0.2
  libyaml = True
```

### DCNM version

* V 12.1.1e

### Affected module(s)

* dcnm_inventory

### Ansible Playbook
```yaml
- name: build inventory of fabric {{ fabric.name }}
  dcnm_inventory:
    fabric: "{{ fabric.name }}"
    config: "{{ updated_fabric.fabric.inventory }}"
  tags:
    - inventory
```

```yaml
# Copy-paste your anisble playbook here 
```

### Expected Behavior

the fabric in playbook does not exist, it could be topo or fabric is not created yet, the error message should indicate it.
### Actual Behavior

Here is the current error message:
```
TASK [fabric : build inventory of fabric fabric_stage] ****************************************************************************************************************************************************************************************
Monday 03 October 2022  09:50:19 -0700 (0:00:00.593)       0:00:02.372 ********
[WARNING]: Persistent connection logging is enabled for 172.25.74.47. This will log ALL interactions to /tmp/ansible.log and WILL NOT redact sensitive configuration like passwords. USE WITH CAUTION!
[...]

Issue #157, opened 2022-08-03
Error on Auth with NDFC


lwy119:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Ansible Version and collection version

2.11.0

### DCNM version

* ND 2.2(1h) NDFC 12.1.1e

### Affected module(s)

* login (cannot login so i think all resources are affected)
* dcnm_interface (just the one I used)

### Ansible Playbook

```yaml
---

- hosts: dcnm_controllers
  gather_facts: false
  connection: ansible.netcommon.httpapi

  collections:
    - cisco.dcnm

  tasks:
    - name: Query Test
      cisco.dcnm.dcnm_interface:
        fabric: fabric
        state: query
        config:
          - switch:
            - "10.1.1.1"
```

### Debug Output

https://gist.github.com/lwy119/e72960dcda164fb34566dc13d8d511c7

### Expected Behavior

Successfully login and perform the query task

### Actual Behavior

Failed at login step

### Steps to Reproduce

ansible-play -i host query.yml

praveenramoorthy:
@lwy119, thanks for raising the issue. Can you please let me know the DCNM collection version which you are using? Can you try with the latest 2.1.0 collection release? This issue has been fixed in that version.

lwy119:
@praveenramoorthy thanks for the reply. Turns out I had a typo in host file. After fixing that the authentication works fine.

However I have another issue. After I created some policies through Ansible, I can see the CLI are pushed to the switches. When I do recalculate and deploy in NDFC, it shows the switches are not in sync and there are pending configs.
If I look into what configs are pending, it basically removes my pushed config and adds them back once….any workaround on this? Is this an expected behavior? It seems NDFC does not recognize configs pushed via Ansible?

dsx1123:
@lwy119 the configuration pushed from policy need match the cli format, especially when you are using freeform template, do you mind share you your playbook here?

innanetguru:
Having a similar issue:

---
- name: Test DCNM 
  hosts: "site_dcnm"
  connection: httpapi
  gather_facts: no 
  collections:
[...]

Issue #140, opened 2022-03-01
maximum recursion depth exceeded error when fabric name does not exist


mikewiebe:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Ansible Version and collection version

```
ansible [core 2.11.8]  (detached HEAD e40051f7c0) last updated 2022/02/10 10:36:01 (GMT -400)
  config file = None
  configured module search path = ['/Users/mwiebe/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/mwiebe/Projects/Ansible/nxos_collection_latest/ansible/lib/ansible
  ansible collection location = /Users/mwiebe/.ansible/collections:/usr/share/ansible/collections
  executable location = /Users/mwiebe/Projects/Ansible/nxos_collection_latest/ansible/bin/ansible
  python version = 3.8.0 (default, Feb 17 2020, 13:58:20) [Clang 11.0.0 (clang-1100.0.33.17)]
  jinja version = 2.11.2
  libyaml = False
```

### DCNM version

* V 12.0.2f

### Affected module(s)

All modules

### Ansible Playbook

```yaml
    - name: MERGED Create L3 network along with all dhcp options
      cisco.dcnm.dcnm_network:
        fabric: fabric_name_does_not_exist
        state: query```
```

### Expected Behavior

Playbook run should fail gracefully with a meaningful error message to indicate that the fabric does not exist.

### Actual Behavior

```
ansible.module_utils.connection.ConnectionError: ['Error on attempt to connect and authenticate with DCNM controller: maximum recursion depth exceeded', 'Error on attempt to connect and authenticate with NDFC controller: maximum recursion depth exceeded', 'Error on attempt to connect and authenticate with NDFC controller: maximum recursion depth exceeded']
```
### Steps to Reproduce

Run the playbook above with an invalid fabric name

mikewiebe:
Fixed in release https://github.com/CiscoDevNet/ansible-dcnm/releases/tag/3.0.0

PR #184, opened 2022-11-30
Changes to add retry limit to send API on 401 return code scenarios

praveenramoorthy:
Fixes issue in #168 

DESCRIPTION:

DCNM return 401 for various scenarios like fabric not present. Current send call to net common retries continuously to login whenever 401 error is returned for a send request, and this happens until session timeout is reached.

FIX:

- Added retry count to send API, now send ill error out once the retry limit is reached. 
- Bumped min required netcommon(send API supports retry count from netcommon version 2.6.1 onwards)

### DIFF
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -13,5 +13,5 @@ description: Ansible collection for the Cisco Nexus® Dashboard Fabric Controlle
 license_file: LICENSE 
 tags: [cisco, ndfc, dcnm, nxos, networking, vxlan]
 dependencies:
-  "ansible.netcommon": "*"
+  "ansible.netcommon": ">=2.6.1"
 repository: https://github.com/CiscoDevNet/ansible-dcnm
--- a/plugins/httpapi/dcnm.py
+++ b/plugins/httpapi/dcnm.py
@@ -42,6 +42,7 @@ class HttpApi(HttpApiBase):
         self.headers = {"Content-Type": "application/json"}
         self.txt_headers = {"Content-Type": "text/plain"}
         self.version = None
+        self.retrycnt = 5 # Retry count for send API
 
     def get_version(self):
         return self.version
@@ -271,6 +272,8 @@ class HttpApi(HttpApiBase):
 
         self.check_url_connection()
 
+        msg = '". Please verify your login credentials, access permissions and fabric details and try again '
+
         try:
             # Perform some very basic path input validation.
             path = str(path)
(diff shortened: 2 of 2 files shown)


mikewiebe:
@praveenramoorthy can you also verify that this fixes the following issues:

https://github.com/CiscoDevNet/ansible-dcnm/issues/157
https://github.com/CiscoDevNet/ansible-dcnm/issues/140

galaxy.yml
```diff
@@ -13,5 +13,5 @@ description: Ansible collection for the Cisco Nexus® Dashboard Fabric Controlle
 license_file: LICENSE 
 tags: [cisco, ndfc, dcnm, nxos, networking, vxlan]
 dependencies:
-  "ansible.netcommon": "*"
+  "ansible.netcommon": ">=2.6.1"
```
mikewiebe, line 16: Have you tested a galaxy install to make sure it installs version `2.6.1` or later?  Just want to make sure this directive works as intended.

plugins/httpapi/dcnm.py
```diff
@@ -42,6 +42,8 @@ def __init__(self, *args, **kwargs):
         self.headers = {"Content-Type": "application/json"}
         self.txt_headers = {"Content-Type": "text/plain"}
         self.version = None
+        # Retry count for send API
+        self.retrycnt = 5
```
mikewiebe, line 46: Minor but personally feel that we should just use `self.retrycount`.  It's only two extra letters and less ambiguous

plugins/httpapi/dcnm.py
```diff
@@ -271,21 +273,26 @@ def send_request(self, method, path, json=None):
...
                 msg = "Value of <path> does not appear to be formated properly"
                 raise ConnectionError(self._return_info(None, method, path, msg))
             response, rdata = self.connection.send(
-                path, json, method=method, headers=self.headers, force_basic_auth=True
+                path, json, self.retrycnt, method=method, headers=self.headers, force_basic_auth=True
             )
             return self._verify_response(response, method, path, rdata)
         except Exception as e:
-            eargs = e.args[0]
+            if e.args:
```
mikewiebe, line 289: Please add a few comments to indicate why this `e.args` check is needed

mikewiebe (changes requested):
Overall it looks good.  Just a few minor comments

plugins/httpapi/dcnm.py
praveenramoorthy: Changes done as suggested

plugins/httpapi/dcnm.py
praveenramoorthy: Changes done as suggested

galaxy.yml
praveenramoorthy: Yes. When I tested for galaxy install, it selected version 4.1.0 for net common installation

PR #190, opened 2023-01-24
FEX interface support initial commit

mmudigon:
The changes in this commit

- support for straight-through FEX interface support
- support for active-active FEX interface support
- related IT and UT cases
- small corrections to existing comments
- related documentation


plugins/modules/dcnm_interface.py
```diff
@@ -1214,6 +1339,184 @@
           admin_state: true                               # Flag to enable/disable Vlan interaface
           mode: vlan                                      # choose from [vlan, vlan_admin_state], default is "vlan"
 
+# AA FEX INTERFACES
+
+- name: Create AA FEX interfaces including optional parameters
```
mikewiebe, line 1344: *nit:* Instead of AA can we use active-active in the name to be more descriptive?

plugins/modules/dcnm_interface.py
```diff
@@ -1214,6 +1339,184 @@
...
+            - no shutdown
+          peer2_cmds:                               # optional, freeform config, default is []
+            - no shutdown
+          admin_state: true                         # Flag to enable/disable FEX interface.
+          enable_netflow: false                     # optional, flag to enable netflow, default is false
+          mode: port_channel_aa                     # choose from [port_channel_aa], default is "port_channel_aa"
+
+# STRAIGHT-THROUGH FEX INTERFACES
+
+- name: Create ST FEX interfaces including optional parameters
```
mikewiebe, line 1441: Here also, let's use straight-through instead of ST

tests/unit/modules/dcnm/test_dcnm_intf.py
```diff
@@ -516,15 +494,6 @@ def load_svi_fixtures(self):
             self.run_dcnm_send.side_effect = [
                 self.playbook_mock_vpc_resp,
                 playbook_svi_intf1,
-                self.playbook_mock_succ_resp,
```
mikewiebe, line 519: Why are these being removed?

mikewiebe (changes requested):

mikewiebe:
@mmudigon Would it be possible to handle the case where the `deploy` flag is not honored for state `delete` as part of these diffs?

plugins/modules/dcnm_interface.py
mmudigon: Sure, will change

plugins/modules/dcnm_interface.py
mmudigon: Sure

tests/unit/modules/dcnm/test_dcnm_intf.py
mmudigon: They are some extra mock data that was added which is not actually required. I have cleaned up some of these and kept only those that are required

mmudigon:
> @mikewiebe 
Sure, I can add the deploy flag check for delete cases as part of this update. Anyway planning to do that once policy changes are committed.

praveenramoorthy (approved):

PR #200, opened 2023-02-22
Uprev ansible supported version

mikewiebe:
We need to upgrade the version of Ansible we support.  I made it consistent with the nxos ansible collection version and removed the upper bound.


mikewiebe:
@praveenramoorthy and @mmudigon This is now ready for review.

praveenramoorthy (approved):

Issue #202, opened 2023-02-24
`dcnm_policy` deploys wrong policy -switch_freefrom


jgomezve:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Ansible Version and collection version

* ansible [core 2.12.3]
### DCNM version

* 12.1.2e

### Affected module(s)

* dcnm_policy

### Ansible Playbook

```yaml
---
- name: Create a Policy in the  Nexus Dashboard Fabric Controller
  hosts: ndfc
  gather_facts: no
  tasks:

  - name: Create policy including required variables
    cisco.dcnm.dcnm_policy:
      fabric: SIMPL-BROWNFIELD
      config:
        - name: switch_freeform               # This must be a valid template name
          description: TESTING_ANSIBLE
          policy_vars:
            CONF: "ip domain-name at"
        - switch:
            - ip: 1.2.3.4
              name: switch_freeform   
              description: TESTING_ANSIBLE
              policy_vars:
                CONF: "ip domain-name at"
```

### Debug Output

```
changed: [NDFC-FRA] => {
    "changed": true,
    "diff": [
        {
            "deleted": [],
            "deploy": [
                {
                    "name": "switch_freeform",
                    "serialNo": "FDO2139044J"
                }
            ],
            "merged": [
                {
                    "description": "TESTING_ANSIBLE",
                    "entityName": "SWITCH",
                    "entityType": "SWITCH",
                    "nvPairs": {
                        "CONF": "ip domain-name at"
                    },
                    "priority": 500,
                    "serialNumber": "FDO2139044J",
                    "source": "",
                    "templateName": "switch_freeform"
                }
            ],
            "query": [],
            "skipped": []
        }
    ],
    "invocation": {
        "module_args": {
            "config": [
                {
                    "description": "TESTING_ANSIBLE",
                    "name": "switch_freeform",
                    "policy_vars": {
                        "CONF": "ip domain-name at"
                    }
                },
                {
                    "switch": [
                        {
                            "description": "TESTING_ANSIBLE",
                            "ip": "10.122.18.114",
                            "name": "switch_freeform",
[...]

PR #203, opened 2023-03-02
Fix to properly process policies under switch block - issue #202

mmudigon:
This commit fixes the following issue (#202) in dcnm_policy module:

Policies specified under the switch block of playbook are not processed as expected. This update fixes this issue.

### DIFF
--- a/tests/integration/targets/dcnm_policy/tests/dcnm/dcnm_policy_merge.yaml
+++ b/tests/integration/targets/dcnm_policy/tests/dcnm/dcnm_policy_merge.yaml
@@ -408,6 +408,74 @@
       loop_control:
         index_var: my_idx
 
+    - name: Delete all created policies
+      cisco.dcnm.dcnm_policy:
+        fabric: "{{ ansible_it_fabric }}"
+        state: deleted                     # only choose form [merged, deleted, query]
+        config:
+          - name: template_101  # This can either be a policy name like POLICY-xxxxx or template name
+          - name: template_102  # This can either be a policy name like POLICY-xxxxx or template name
+          - name: template_103  # This can either be a policy name like POLICY-xxxxx or template name
+          - name: template_104  # This can either be a policy name like POLICY-xxxxx or template name
+          - name: template_105  # This can either be a policy name like POLICY-xxxxx or template name
+          - switch:
+             - ip: "{{ ansible_switch1 }}"
+      register: result
+
+    - assert:
+        that:
+          - 'item["RETURN_CODE"] == 200'
+          - 'item["MESSAGE"] == "OK"'
+      loop: '{{ result.response }}'
+
+
+    - name: Create policies - checking for policies under switch block
+      cisco.dcnm.dcnm_policy:
+        fabric: "{{ ansible_it_fabric }}"
+        deploy: false
+        config:
+          - name: template_101  # This must be a valid template name
+            description: create template_101
+            priority: 101
+
+          - name: template_102  # This must be a valid template name
+            description: create template_102
+            priority: 102
+
+          - name: template_103  # This must be a valid template name
+            description: create template_103
+            priority: 103
+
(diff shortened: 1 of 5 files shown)


mikewiebe:
Just making sure I understand the expected behavior.

For the following playbook:

```yaml
  - name: Create policy including required variables
    cisco.dcnm.dcnm_policy:
      fabric: SIMPL-BROWNFIELD
      config:
        - name: switch_freeform               # This must be a valid template name
          description: TESTING_ANSIBLE
          policy_vars:
            CONF: "ip domain-name at"
        - switch:
            - ip: 1.2.3.4
              name: switch_freeform   
              description: TESTING_ANSIBLE
              policy_vars:
                CONF: "ip domain-name at"
```

The parameters directly under config would apply to all switches unless there is a more specific set of parameters under a specific switch IP.  Essentially I would expect the parameters under config to be ignored for switch with ip `1.2.3.4` and instead use the more specific parameters under the switch section.

Is that correct?

tests/integration/targets/dcnm_policy/tests/dcnm/dcnm_policy_merge.yaml
```diff
@@ -408,6 +408,74 @@
...
+                  - name: template_104  # This must be a valid template name
+                    description: create template_104 - with changes
+                    priority: 104
+
+                  - name: template_105  # This must be a valid template name
+                    description: create template_105 - with changes
+                    priority: 105
+      register: result
+
+    - assert:
```
mikewiebe, line 471: For this test case is the following accurate:

* Policy using `template_101` is created on `ansible_switch1`
* Policy using `template_102` is created on `ansible_switch1`
* Policy using `template_103` is cretaed on `ansible_switch1` and `ansible_switch2`
* Policy using `teamplte_104` is created on `ansible_swtich2`
* Policy using `teamplte_105` is created on `ansible_swtich2`

If this is the case can we check to make sure that the policies were deployed on the correct switch ip's in the assert block?

mmudigon: Sure let me add the checks. The only thing we need to be aware of is that the response only has serial numbers, and so we need to add these variables to group_vars for us to check. Otherwise this test case may fail. In case we plan to setup dynamic topologies and add random switches, we need to ensure that proper serial numbers are also defined for these in the variables.

mmudigon: Regarding the scope of policies, I have interpreted in one way. May be we should discuss this in our meeting on 7th Mar, to finalise the interpretation. The complexity arises in case all policies are same i.e. under config as well as the switch block. We will discuss this and finalise.

mikewiebe:
@mmudigon Can you take a look at https://github.com/CiscoDevNet/ansible-dcnm/issues/212 and address as part of these changes?

mmudigon:
Hi Mike,

Asked Shangxin to verify this with the latest fixes we made to Policy module (#203). He said he did not see the issue with latest fixes which is still in PR #203

Thanks

Regards
Mallik

From: Mike Wiebe ***@***.***>
Date: Thursday, 16 March 2023 at 20:49
To: CiscoDevNet/ansible-dcnm ***@***.***>
Cc: MALLIK MUDIGONDA (mmudigon) ***@***.***>, Mention ***@***.***>
Subject: Re: [CiscoDevNet/ansible-dcnm] Fix to properly process policies under switch block - issue #202 (PR #203)

@mmudigon<https://github.com/mmudigon> Can you take a look at #212<https://github.com/CiscoDevNet/ansible-dcnm/issues/212> and address as part of these changes?

mikewiebe:
@mmudigon Are the changes for this PR completed or is there still something pending here?

mikewiebe (approved):

Issue #197, opened 2023-02-11
NDFC: Error when using the dcnm_network.py module in ansible_collections/cisco/dcnm/plugins/modules


smukkavi:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### DCNM version

* V 15.1.1e

### Affected module(s)

dcnm_network.py 

### Ansible Playbook

```yaml
# Copy-paste your anisble playbook here 
```

### Expected Behavior

Customer is using ingress replication for VxLAN setup in their test lab, this option should not expect a multicast address. But while running the Ansible playbook, they see error message about the multicast address which is not set anywhere.

### Actual Behavior

We are getting an error when using the dcnm_network.py module in ansible_collections/cisco/dcnm/plugins/modules.
 
TASK [configure_network : Merge networks] ******************************************************************************************************************
[...]

Issue #185, opened 2023-01-05
v2.4.0 dcnm_network throws errors with unicast-only fabrics


allenrobel:
After upgrade from v2.3.0 to v2.4.0, either of the following errors may be seen with cisco.dcnm.dcnm_network.

If network(s) do not already exist:

```
"Invalid: Fabric mode is not multicast and Multicast Address: 239.1.1.0 is present"
```

Or, if networks already exist and ``replace`` is used:

```
"Invalid: Ingress Replication is true and Multicast Address is 239.1.1.0"
```

Below is an example playbook to recreate this:

```yaml
- hosts: ndfc
  gather_facts: false
  name: test
  tasks:
  - cisco.dcnm.dcnm_network:
     fabric: MSD
     state: replaced
     config:
     - net_name: n1111
       vrf_name: v1
       vlan_id: 1111
       gw_ip_subnet: 10.21.1.1/24
       attach:
  vars:
    ansible_connection: httpapi
```

I believe the following diff between v2.3.0 and v2.4.0 is relevant here:

```
2175c2187
<                 multicast_group_address=dict(type="ipv4", default=""),
---
>                 multicast_group_address=dict(type="ipv4", default="239.1.1.0"),
2239c2251
<                 multicast_group_address=dict(type="ipv4", default=""),
---
>                 multicast_group_address=dict(type="ipv4", default="239.1.1.0"),
```

A workaround is to add ``multicast_group_address`` with an empy string value, per below:

```yaml
- hosts: ndfc
  gather_facts: false
  name: test
  tasks:
  - cisco.dcnm.dcnm_network:
     fabric: MSD
     state: replaced
     config:
     - net_name: n1111
       multicast_group_address: ""
       vrf_name: v1
       vlan_id: 1111
       gw_ip_subnet: 10.21.1.1/24
       attach:
  vars:
    ansible_connection: httpapi
```

praveenramoorthy:
@allenrobel Thanks for raising this issue. We will check and will work on a fix.

mikewiebe:
Quick update here.  I was able to recreate the issue and as stated by @allenrobel it's specific to unicast-only fabrics.  Working on a fix

praveenramoorthy:
Fixed as part of #208

Issue #207, opened 2023-03-09
Module dcnm_vrf clears unsupported VRF Template properties


jgomezve:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Ansible Version and collection version

* 2.12.3

### DCNM version

* V 12.1.2e

### Affected module(s)

* dcnm_vrf

### Ansible Playbook

```yaml
  - name: Merge vrfs
    cisco.dcnm.dcnm_vrf:
      fabric: myFabric
      state: merged
      config:
      - vrf_name: VRF-TEST
        vrf_id: 10010
        vrf_template: myVrfTemplate
        vrf_extension_template: myVrfExtensionTemplate
        vlan_id: 2010
        service_vrf_template: null
        attach:
        - ip_address: 1.2.3.4
          deploy: true

```

### Debug Output

The module `dcnm_rest` is being used to create/update a VRF, and then the modules `dcnm_vrf` to deploy/ attaches to the switches

The playbook is executed without any problem however most of the VRF attributes, the ones not supported by the module, are cleared

### Expected Behavior

Properties not supported by the module should be left untouched

### Actual Behavior

Properties such as MTU, Tag, Max BGP paths etc are cleard from the VRF config

### Steps to Reproduce

* Create a VRF manually in NDFC or using the `dcnm_rest` module
* Execute the playbook
* Attributes not supported by the modules are cleared

### References

The root cause seem to be a "limitation" in the NDFC Rest API. I have manually create a VRF and then updated it via Postman with the following payload

```json
{
    "displayName": "VRF-TEST",
    "fabric": "myFabric",
    "vrfExtensionTemplate": "myVrfExtensionTemplate",
    "vrfId": 10011,
    "vrfName": "VRF-TEST"",
    "vrfTemplate": "myVrfTemplate",
    "vrfTemplateConfig": "{\"vrfName\":\"VRF-TEST\",\"vrfVlanId\":2010,\"vrfSegmentId\":10011,\"nveId\":\"1\",\"asn\":\"\"}"
}
```

[...]

PR #208, opened 2023-03-10
Support for all network and vrf configs

praveenramoorthy:
Added support for all network and vrf configs

Fix for issue in #185 and #197


tests/unit/modules/dcnm/fixtures/dcnm_net.json
```diff
@@ -0,0 +1,555 @@
+{
```
mikewiebe, line 1: Minor note but to be consistent can we use the full name for this file `dcnm_network`?

tests/unit/modules/dcnm/test_dcnm_net.py
```diff
@@ -0,0 +1,737 @@
+# Copyright (c) 2020-2023 Cisco and/or its affiliates.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
```
mikewiebe, line 5: Same here to be consistent can we use the full name for this file `test_dcnm_network.py`?

tests/unit/modules/dcnm/test_dcnm_net.py
```diff
@@ -0,0 +1,737 @@
+# Copyright (c) 2020-2023 Cisco and/or its affiliates.
```
mikewiebe, line 1: Since these are new files we can just use 2023 in the date.

tests/unit/modules/dcnm/fixtures/dcnm_net.json
praveenramoorthy: Changes done as suggested.

tests/unit/modules/dcnm/test_dcnm_net.py
praveenramoorthy: Changes done as suggested.

tests/unit/modules/dcnm/test_dcnm_net.py
praveenramoorthy: Changes done as suggested.

mikewiebe:
@praveenramoorthy This issue was raised https://github.com/CiscoDevNet/ansible-dcnm/issues/207.  I am wondering if the updates in this PR prevent this from happening.  Can you verify?

plugins/modules/dcnm_network.py
```diff
@@ -2223,6 +2533,11 @@ def validate_input(self):
 
         else:
 
+            if self.dcnm_version > 11:
```
mikewiebe, line 2536: Please add a comment indicating why this is needed to avoid future confusion

tests/integration/targets/dcnm_network/tests/dcnm/replaced.yaml
```diff
@@ -115,14 +115,14 @@
...
-- name: Query fabric state until networkStatus transitions to DEPLOYED state
+- name: Query fabric state until networkStatus transitions to NA state
   cisco.dcnm.dcnm_network:
     fabric: "{{ test_fabric }}"
     state: query
   register: query_result
   until:
-    - "query_result.response[0].parent.networkStatus is search('DEPLOYED')"
-    - "query_result.response[1].parent.networkStatus is search('DEPLOYED')"
+    - "query_result.response[0].parent.networkStatus is search('NA')"
```
mikewiebe, line 124: Why are we changing this from `DEPLOYED` to `NA`?

tests/integration/targets/dcnm_vrf/tests/dcnm/merged.yaml
```diff
@@ -65,6 +65,10 @@
     - '"{{ ansible_switch2 }}" in result.diff[0].attach[1].ip_address'
     - 'result.diff[0].vrf_name == "ansible-vrf-int1"'
 
+- name: MERGED - sleep for 40 seconds for DCNM to completely update the state
```
mikewiebe, line 68: Here and elsewhere, let's try and avoid arbitrary sleep statements like this and instead use the following approach:

https://github.com/CiscoDevNet/ansible-dcnm/blob/develop/tests/integration/targets/dcnm_network/tests/dcnm/self-contained-tests/sm_mcast_params.yaml#L32

mikewiebe: Better to check for a specific condition and loop until that condition is reached instead of hoping 40 seconds is long enough.  Also if we reach the state sooner then we move to the next task more quickly.

mikewiebe (changes requested):

plugins/modules/dcnm_vrf.py
```diff
@@ -84,6 +84,208 @@
         - Service vrf template
         type: str
         default: None
+      vrf_vlan_name:
+        description:
+        - VRF Vlan Name
+        - if > 32 chars enable:system vlan long-name
```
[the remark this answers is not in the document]
praveenramoorthy: I did in another place and I had to remove an example with IPv6 address. But for some reason to did not error here. I will changed ':' to '-'.

tests/integration/targets/dcnm_network/tests/dcnm/replaced.yaml
praveenramoorthy: In this particular test we are  not attaching any switches to network, so the switch detail will never have status as deployed. It will be be just NA.

plugins/modules/dcnm_network.py
```diff
@@ -181,7 +246,7 @@
           ports:
             description:
             - List of switch interfaces where the network will be attached
-            type: str
+            type: list
```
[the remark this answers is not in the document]
praveenramoorthy: Changes done as suggested

plugins/modules/dcnm_network.py
praveenramoorthy: Changes done as suggested

plugins/modules/dcnm_network.py
```diff
@@ -418,7 +483,7 @@ class DcnmNetwork:
             "GET_NET_ID": "/rest/managed-pool/fabrics/{}/segments/ids",
             "GET_NET": "/rest/top-down/fabrics/{}/networks",
             "GET_NET_NAME": "/rest/top-down/fabrics/{}/networks/{}",
-            "GET_VLAN": "/rest/resource-manager/vlan/{}?vlanUsageType=TOP_DOWN_VRF_VLAN",
+            "GET_VLAN": "/rest/resource-manager/vlan/{}?vlanUsageType=TOP_DOWN_NETWORK_VLAN",
```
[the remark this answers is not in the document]
praveenramoorthy: :)

plugins/modules/dcnm_network.py
praveenramoorthy: Changes done as suggested

tests/integration/targets/dcnm_vrf/tests/dcnm/merged.yaml
praveenramoorthy: Changes done as suggested

mikewiebe (approved):

PR #215, opened 2023-03-22
Fix for config-save with pre-provision switches only

praveenramoorthy:
This commit fixes the config-save issue when there are only pre-provision switches in playbook along with only pre-provision switches in fabric.


praveenramoorthy:
> Does this fix address one of the open issues?

No. Found this out during a debug session for some other issue.

PR #219, opened 2023-04-21
Vrf lite enhancements

praveenramoorthy:
- Added support to configure multiple vrf lite extensions
- Added idempotence support for vrf lite
- IT/UT cases for new support

### DIFF
--- a/plugins/action/dcnm_vrf.py
+++ b/plugins/action/dcnm_vrf.py
@@ -52,7 +52,7 @@ class ActionModule(ActionNetworkModule):
                                         )
                                         display.warning(msg)
                             except TypeError:
-                                msg = "Please specifiy atleast one VRF LITE parameter in attach"
+                                msg = "Please specifiy at least one VRF LITE parameter in attach"
                                 return {"failed": True, "msg": msg}
 
         self.result = super(ActionModule, self).run(task_vars=task_vars)
--- a/plugins/action/dcnm_vrf.py
+++ b/plugins/action/dcnm_vrf.py
@@ -38,6 +38,22 @@ class ActionModule(ActionNetworkModule):
                         if "vlan_id" in at:
                             msg = "Playbook parameter vlan_id should not be specified under the attach: block. Please specify this under the config: block instead"  # noqa
                             return {"failed": True, "msg": msg}
+                        if "vrf_lite" in at:
+                            try:
+                                for vl in at["vrf_lite"]:
+                                    if (
+                                        not vl.get("interface")
+                                    ):
+                                        msg = (
+                                                "'interface' parameter is not specified in playbook under 'vrf_lite' config. "
+                                                "Idempotence check will happen based on first available best match. "
+                                                "While attaching, first available extension will be attached for VRF LITE. "
+                                                "For best results specify all possible vrf_lite parameters."
+                                              )
+                                        display.warning(msg)
+                            except TypeError:
+                                msg = "Please specifiy atleast one VRF LITE parameter in attach"
+                                return {"failed": True, "msg": msg}
 
         self.result = super(ActionModule, self).run(task_vars=task_vars)
         return self.result
--- a/plugins/action/dcnm_vrf.py
+++ b/plugins/action/dcnm_vrf.py
@@ -45,11 +45,11 @@ class ActionModule(ActionNetworkModule):
                                         not vl.get("interface")
                                     ):
                                         msg = (
-                                                "'interface' parameter is not specified in playbook under 'vrf_lite' config. "
-                                                "Idempotence check will happen based on first available best match. "
-                                                "While attaching, first available extension will be attached for VRF LITE. "
-                                                "For best results specify all possible vrf_lite parameters."
-                                              )
+                                            "'interface' parameter is not specified in playbook under 'vrf_lite' config. "
+                                            "Idempotence check will happen based on first available best match. "
+                                            "While attaching, first available extension will be attached for VRF LITE. "
+                                            "For best results specify all possible vrf_lite parameters."
+                                        )
                                         display.warning(msg)
                             except TypeError:
                                 msg = "Please specifiy atleast one VRF LITE parameter in attach"
--- a/plugins/modules/dcnm_vrf.py
+++ b/plugins/modules/dcnm_vrf.py
@@ -449,6 +449,8 @@ EXAMPLES = """
       - ip_address: 192.168.1.225
         vrf_lite:
          # All parameters under vrf_lite are optional
+         # For best results in VRF LITE idempotence checks and
+         # configurations specify all possible vrf_lite parameters
           - peer_vrf: test_vrf_1 # optional
             interface: Ethernet1/16 # optional
             ipv4_addr: 10.33.0.2/30 # optional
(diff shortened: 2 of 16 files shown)


plugins/action/dcnm_vrf.py
```diff
@@ -38,6 +38,22 @@ def run(self, tmp=None, task_vars=None):
...
+                                    ):
+                                        msg = (
+                                            "'interface' parameter is not specified in playbook under 'vrf_lite' config. "
+                                            "Idempotence check will happen based on first available best match. "
+                                            "While attaching, first available extension will be attached for VRF LITE. "
+                                            "For best results specify all possible vrf_lite parameters."
+                                        )
+                                        display.warning(msg)
+                            except TypeError:
+                                msg = "Please specifiy atleast one VRF LITE parameter in attach"
```
mikewiebe, line 55: `/atleast/at least`

plugins/modules/dcnm_vrf.py
```diff
@@ -448,15 +448,21 @@
       - ip_address: 192.168.1.224
       - ip_address: 192.168.1.225
         vrf_lite:
-         # All parameters under vrf_lite except peer_vrf are optional and
-         # will be supplied by DCNM when omitted in the playbook
-          - peer_vrf: test_vrf_1 # peer_vrf is mandatory
+         # All parameters under vrf_lite are optional
```
mikewiebe, line 451: Should you expand this comment to indicate that while they are all optional specifying more parameters will result in a more accurate match?  Or at least guidance on a best practice here?

tests/integration/targets/dcnm_vrf/tests/dcnm/merged.yaml
```diff
@@ -204,7 +204,14 @@
     fabric: "{{ test_fabric }}"
     state: deleted
 
-- name: MERGED - Create, Attach and Deploy new VRF - VRF/VRF LITE EXTENSION Provided by the DCNM - Only Mandatory option - Rest are populated from DCNM
+- name: QUERY - sleep for 40 seconds for DCNM to completely remove lite profile
```
mikewiebe, line 207: Can we update the tests to use `retry` instead similar to https://github.com/CiscoDevNet/ansible-dcnm/blob/develop/tests/integration/targets/dcnm_network/tests/dcnm/self-contained-tests/merged_net_all.yaml#L36

Let's try to remove arbitrary sleep whenever possible

tests/integration/targets/dcnm_vrf/tests/dcnm/merged.yaml
```diff
@@ -310,15 +317,14 @@
...
-          - interface: "{{ ansible_int1 }}"
       deploy: true
   register: result
   ignore_errors: yes
 
 - assert:
     that:
     - 'result.changed == false'
-    - '"peer_vrf is mandatory under attach VRF LITE parameters" in result.msg'
+    - '"Please specifiy atleast one VRF LITE parameter in attach" in result.msg'
```
mikewiebe, line 327: `/atleast/at least` After you change the code :)

tests/integration/targets/dcnm_vrf/tests/dcnm/self-contained-tests/vrf_lite.yaml
```diff
@@ -0,0 +1,286 @@
...
+- assert:
+    that:
+    - 'result.response.DATA != None'
+
+- name: MERGED - Clean up any existing vrfs
+  cisco.dcnm.dcnm_vrf:
+    fabric: "{{ test_fabric }}"
+    state: deleted
+
+- name: VRF LITE - sleep for 40 seconds for DCNM to completely update the state
```
mikewiebe, line 28: Use `retry` block

https://github.com/CiscoDevNet/ansible-dcnm/blob/develop/tests/integration/targets/dcnm_vrf/tests/dcnm/merged.yaml#L50

tests/integration/targets/dcnm_vrf/tests/dcnm/self-contained-tests/vrf_lite.yaml
```diff
@@ -0,0 +1,286 @@
...
+  cisco.dcnm.dcnm_vrf: &conf1
+    fabric: "{{ test_fabric }}"
+    state: merged
+    config:
+    - vrf_name: ansible-vrf-int1
+      vrf_id: 9008011
+      vrf_template: Default_VRF_Universal
+      vrf_extension_template: Default_VRF_Extension_Universal
+      vlan_id: 500
+      source: null
```
mikewiebe, line 46: Do we ever set `source` or `service_vrf_template` properties?

tests/integration/targets/dcnm_vrf/tests/dcnm/self-contained-tests/vrf_lite.yaml
```diff
@@ -0,0 +1,286 @@
...
+  cisco.dcnm.dcnm_vrf:
+    fabric: "{{ test_fabric }}"
+    state: query
+  register: query_result
+  until:
+    - "query_result.response[0].parent.vrfStatus is search('DEPLOYED')"
+  retries: 30
+  delay: 2
+
+- assert:
```
mikewiebe, line 72: Are we checking for vrf-lite specifically in this assert statement?  Based on the task above I would expect vrf-lite to only be associated with `switch2` and not `switch1`.  Is that correct?  We should try and explicitly check for vrf-lite config/behavior in the assert here and in other tests in the file

tests/integration/targets/dcnm_vrf/tests/dcnm/self-contained-tests/vrf_lite.yaml
```diff
@@ -0,0 +1,286 @@
...
+    - vrf_name: ansible-vrf-int1
+      vrf_id: 9008011
+      vrf_template: Default_VRF_Universal
+      vrf_extension_template: Default_VRF_Extension_Universal
+      vlan_id: 500
+      source: null
+      service_vrf_template: null
+      attach:
+      - ip_address: "{{ ansible_switch1 }}"
+      - ip_address: "{{ ansible_switch2 }}"
```
mikewiebe, line 171: Should we also be testing vrf-lite under more then one switch?

plugins/action/dcnm_vrf.py
praveenramoorthy: Changes done as suggested

plugins/modules/dcnm_vrf.py
praveenramoorthy: Changes done as suggested

tests/integration/targets/dcnm_vrf/tests/dcnm/merged.yaml
praveenramoorthy: This can be removed after check_deploy support, which is planned for a later release.

tests/integration/targets/dcnm_vrf/tests/dcnm/self-contained-tests/vrf_lite.yaml
praveenramoorthy: This can be removed after check_deploy support, which is planned for a later release.

tests/integration/targets/dcnm_vrf/tests/dcnm/merged.yaml
praveenramoorthy: Changes done as suggested

tests/integration/targets/dcnm_vrf/tests/dcnm/self-contained-tests/vrf_lite.yaml
praveenramoorthy: No we don't. I have removed both these parameter from all the plays

tests/integration/targets/dcnm_vrf/tests/dcnm/self-contained-tests/vrf_lite.yaml
praveenramoorthy: Changes done to check for vrf-lite configurations for the associated switch.

Issue #216, opened 2023-04-14
dcnm_links: src_interface / dst_interface case sensitivity for state == query


allenrobel:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

This could probably be fixed on the NX-OS side, but submitting this to see if it could be addressed within the dcnm_links module by ignoring upper/lower-case when returning results.

The problem:

N9K-C9504 reports its loopback as: loopback0
N9K-C9336C-FX2 reports its loopback as: Loopback0

If the user runs the below task, it will return no result since both interfaces are specified with "loopback0" (below, dst_interface belongs to N9K-C9336C-FX2):

```yaml
    - name: Query RS - BGW peering loopbacks
      cisco.dcnm.dcnm_links:
        state: query
        src_fabric: "EF_1"
        config:
          - dst_fabric: "DCI"
            src_interface: "loopback0"
            dst_interface: "loopback0"
```

In contrast, the below task returns a result:

```yaml
    - name: Query RS - BGW peering loopbacks
      cisco.dcnm.dcnm_links:
        state: query
        src_fabric: "EF_1"
        config:
          - dst_fabric: "DCI"
            src_interface: "loopback0"
            dst_interface: "Loopback0"
```

Hence, the user has to know what switch models are deployed in order to obtain correct results.  It would be wonderful if dcnm_links could ignore ``if-name`` case (see JSON below) when determining whether to include a link in the result.

For reference, below is the relevant output from the second playbook above (removing non-relevant key/values for brevity).

```json
{
    "result": {
        "response": [
            {
                "fabricName": "EF_1<->DCI",
                "nvPairs": {
                    "DEST_FABRIC": "DCI",
                    "SOURCE_FABRIC": "EF_1",
                },
                "policyId": "POLICY-57110",
                "sw1-info": {
                    "fabric-name": "EF_1",
                    "if-name": "loopback0",
                    "sw-model-name": "N9K-C9504",
                    "switch-role": "border gateway",
                },
                "sw2-info": {
                    "fabric-name": "DCI",
                    "if-name": "Loopback0",
[...]

PR #220, opened 2023-04-24
Added checks for fabric monitoring status and switch's managable attributes during deploy

mmudigon:
This commit includes changes to dcnm_links module to:
     - identify all switches which are in monitoring state
     - identify all switches which are managable
     - check for monitoring and managable attributes before deploy
     - avoid deploying on fabrics which are in monitoring state
     - avoid deploy to switches which have managable set to false

Added IT and UT cases to verify the above scenarios.
Updated the dcnm.py module to add logicalName to Serial Number mappings to inventory_data in case ipAddress is "". This will be the case for META switches.

### DIFF
--- a/tests/integration/targets/dcnm_links/tests/dcnm/dcnm_links_misc.yaml
+++ b/tests/integration/targets/dcnm_links/tests/dcnm/dcnm_links_misc.yaml
@@ -1,0 +1,900 @@
+##############################################
+##               SETUP                      ##
+##############################################
+
+- name: Remove local log file
+  local_action: command rm -f dcnm_links.log
+
+- block:
+
+#############################################
+#               DELETE                     ##
+#############################################
+
+    - name: Initial setup - Delete Links on numbered fabric
+      cisco.dcnm.dcnm_links: &links_delete
+        state: deleted                                           # choose from [merged, replaced, deleted, query]
+        src_fabric: "{{ ansible_num_fabric }}"
+        config:
+          - dst_fabric: "{{ ansible_num_fabric }}"               # Destination fabric
+            src_interface: "{{ intf_1_3 }}"                      # Interface on the Source fabric
+            dst_interface: "{{ intf_1_3 }}"                      # Interface on the Destination fabric
+            src_device: "dummy-switch-1"                         # Device on the Source fabric
+            dst_device: "{{ ansible_num_switch1 }}"              # Device on the Destination fabric
+            template: ext_fabric_setup                           # template to be applied, choose from 
+                                                                 #   [ ext_fabric_setup, ext_multisite_underlay_setup,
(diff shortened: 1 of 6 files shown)


mikewiebe:
@mmudigon Could you check and see if it would be possible to fix this issue (https://github.com/CiscoDevNet/ansible-dcnm/issues/216) as part of this PR?  Looks like it might be forward and since you are already modifying the `dcnm_links` modules?

mmudigon:
I am still checking if ignoring case is ok. The issue maybe for interface names, but should look at other fields as well. If it is a simple change I can include that.

Regards
Mallik

From: Mike Wiebe ***@***.***>
Date: Wednesday, 26 April 2023 at 19:27
To: CiscoDevNet/ansible-dcnm ***@***.***>
Cc: MALLIK MUDIGONDA (mmudigon) ***@***.***>, Mention ***@***.***>
Subject: Re: [CiscoDevNet/ansible-dcnm] Added checks for fabric monitoring status and switch's managable attributes during deploy (PR #220)

@mmudigon<https://github.com/mmudigon> Could you check and see if it would be possible to fix this issue (#216<https://github.com/CiscoDevNet/ansible-dcnm/issues/216>) as part of this PR? Looks like it might be forward and since you are already modifying the dcnm_links modules?

tests/integration/targets/dcnm_links/tests/dcnm/dcnm_links_misc.yaml
```diff
@@ -0,0 +1,900 @@
...
+#############################################
+#               DELETE                     ##
+#############################################
+
+    - name: Initial setup - Delete Links on numbered fabric
+      cisco.dcnm.dcnm_links: &links_delete
+        state: deleted                                           # choose from [merged, replaced, deleted, query]
+        src_fabric: "{{ ansible_num_fabric }}"
+        config:
+          - dst_fabric: "{{ ansible_num_fabric }}"               # Destination fabric
```
mikewiebe, line 19: Remind me, does this module delete all links if state is deleted and we omit the config block from the playbook?

mikewiebe (commented):
Changes look fine to me.  Just one question/comment.

tests/integration/targets/dcnm_links/tests/dcnm/dcnm_links_misc.yaml
mmudigon: No, config block is mandatory for Links Delete. This is because the system by default has links when it comes up and if the user by mistake executes a delete with no config block, then it will result in all links being deleted which may not be desirable. So to avoid this we force the user to explicitly include the config block so that only the matching links are deleted

mikewiebe (approved):

PR #224, opened 2023-05-24
Fix default mcast group address in network module

praveenramoorthy:
The default multicast group address should be "" in case of networks creates in fabrics with ingress replication enabled.


praveenramoorthy:
> Looks good but can we add a test to cover this?

Added IT cases

PR #225, opened 2023-06-09
Fix to check for deploy flag during delete operation

mmudigon:
This commit in dcnm_interface module includes the following updates:
    - Check deploy flag during deleted and overridden states. If deploy is false, do not deploy deleted interfaces. Just mark them as deleted.
    - Check if switches included in playbook are manageable. If not return error indicating the same
    - Check if Fabric mode is read_only. If it is read_only return error without making any changes

### DIFF
--- a/plugins/modules/dcnm_interface.py
+++ b/plugins/modules/dcnm_interface.py
@@ -55,6 +55,15 @@ options:
       - deleted
       - query
     default: merged
+  deploy:
+    description:
+    - Flag indicating if the configuration must be pushed to the switch. This flag is used to decide the deploy behavior in
+      'deleted' and 'overridden' states as mentioned below
+    - In 'overridden' state this flag will be used to deploy deleted interfaces.
+    - In 'deleted' state this flag will be used to deploy deleted interfaces when a specific 'config' block is not
+      included.
+    type: bool
+    default: true
   config:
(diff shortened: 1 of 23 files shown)


plugins/modules/dcnm_interface.py
```diff
@@ -55,6 +55,15 @@
       - deleted
       - query
     default: merged
+  deploy:
```
mikewiebe, line 58: I am a bit confused here.  Why are we adding a global level `deploy` flag that seems to only control the case where we are marking things for delete?  I thought that we were going to reuse the `deploy` flag under the `config` section to control both create and delete cases.

For example using loopback:

* Flag `deploy` under `config` section controls if loopback config to add/remove loopback is sent to device

mmudigon: We have introduced this global flag specifically to deal with deleted and overridden states. In the case of delete, there is an option to specify deleted state without including any config block. In this case there is no way to identify if user really wants to deploy after performing delete. Similarly in the case of overridden state, the config block includes a deploy flag which is applicable to the new config included. But the user has no way to specify if the deleted/defaulted information should be deployed or not. This flag handles these special cases.

mikewiebe (approved):

Issue #242, opened 2023-08-10
NDFC Support for login Domain


samurato:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description
Provide a way to define login domain that can be used to point with  the TACAS or local NDFC

### New or Affected modules(s):
All existing :- Handles Authentication

* dcnm_XXXXX

### DCNM version

* V NDFC:- 12.1.2e, with ND 2.3.2d

### Potential ansible task config

```yaml
# Copy-paste your ansible playbook
all:
  vars:
    ansible_user: "dcnm_username"
    ansible_password: "dcnm_password"
    ansible_domain: ISE <<<<<<<<<<<<<<<<<<<
    ansible_python_interpreter: python
    ansible_httpapi_validate_certs: False
    ansible_httpapi_use_ssl: True
  children:
    dcnm_controllers:
      hosts:
        192.168.2.10:
           ansible_network_os: cisco.dcnm.dcnm
```

### References

**Additional context**

mikewiebe:
Fixed in the latest release

PR #243, opened 2023-08-10
add login domain support for httpapi plugin

dsx1123:
#242


mikewiebe (commented):
LGTM overall.  Just want to run a few local tests on my setup to double check a few workflows and then will merge so we can cut a new release next week.

Issue #248, opened 2023-08-16
support two new types of link template `ext_vxlan_mpls_overlay_setup` and `ext_vxlan_mpls_underlay_setup`


dsx1123:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

These two types of link templates are needed for mpls hand off

### New or Affected modules(s):

* dcnm_link

### DCNM version

* V 12.1.2e

### Potential ansible task config

```yaml
# Copy-paste your ansible playbook
```

### References

**Additional context**

Issue #236, opened 2023-08-01
dcnm_link error when trying to modify the auto-discovered link


dsx1123:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Ansible Version and collection version

```
ansible [core 2.12.6]
  config file = /home/shdu/.ansible.cfg
  configured module search path = ['/home/shdu/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/shdu/virtualenv/ansible/lib/python3.11/site-packages/ansible
  ansible collection location = /home/shdu/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/shdu/virtualenv/ansible/bin/ansible
  python version = 3.11.2 (main, Mar 13 2023, 12:18:29) [GCC 12.2.0]
  jinja version = 3.1.2
  libyaml = True

```

### DCNM version

* V 12.1.2e

### Affected module(s)

* dcnm_link

### Ansible Playbook

```yaml
# Copy-paste your anisble playbook here 
   - name: modify links between leaf and spine
      cisco.dcnm.dcnm_links:
        src_fabric: "{{ fabric.name }}"
        config:
          - dst_fabric: "{{ fabric.name }}"
            src_device: "{{ fabric.leaf[0] }}"
            dst_device: "{{ fabric.spine[0] }}"
            src_interface: "Ethernet1/5"
            dst_interface: "Ethernet1/1"
            template: int_intra_fabric_num_link
            profile:
              peer1_ipv4_addr: 10.30.1.1
              peer2_ipv4_addr: 10.30.1.2
              peer1_ipv6_addr: ""
              peer2_ipv6_addr: ""
              admin_state: true
              mtu: 9216

```

### Debug Output

https://gist.github.com/dsx1123/764737ed309378de23f41bb0e6ae62c8
### Steps to Reproduce

this link is auto-discovered during adding switches. so the template name is empty which cause the bug.

dsx1123:
@mmudigon  could you please take a look?

mmudigon:
Hi Shangxin,

[...]

PR #253, opened 2023-08-22
dcnm_link issues

dsx1123:
- fix an issue when the link is already discovered by NDFC #236 
- with the fix above, issue #244 should be eased as the source and destination fabric/device/interface should be determined. A more bulletproof fix should switch the source and destination and then compare what is configured already and what is defined in the playbook but will need more effort, will revisit it later.
- Adding two types of links. ext_vxlan_mpls_overlay_setup and ext_vxlan_mpls_underlay_setup #248

### DIFF
--- a/docs/cisco.dcnm.dcnm_links_module.rst
+++ b/docs/cisco.dcnm.dcnm_links_module.rst
@@ -182,6 +182,47 @@ Parameters
                     <td class="elbow-placeholder"></td>
                 <td colspan="1">
                     <div class="ansibleOptionAnchor" id="parameter-"></div>
+                    <b>dci_routing_proto</b>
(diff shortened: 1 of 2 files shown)


mmudigon (approved):

mikewiebe (commented):
Changes look fine but should we add integration and unit tests to cover the updates.

dsx1123:
> Changes look fine but should we add integration and unit tests to cover the updates.

integration testing for the change will be hard, for the first change, need to make sure there is an auto-discovered link, for the second change, it requires physical devices.

I will add a unit test for the templateName is None scenario.

mikewiebe (approved):

Issue #245, opened 2023-08-11
support secondary_ip of loopback interface


dsx1123:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### New or Affected modules(s):

* dcnm_interface

### NDFC version

* V 12.1.2e

### Potential ansible task config

```yaml
# Copy-paste your ansible playbook
- name: Create loopback interfaces
  cisco.dcnm.dcnm_interface: 
    fabric: mmudigon-fabric
    state: merged                         # only choose from [merged, replaced, deleted, overridden, query]
    config:
      - name: lo100                       # should be of the form lo<port-id>
        type: lo                          # choose from this list [pc, vpc, sub_int, lo, eth]
        switch:
          - "192.172.1.1"                 # provide the switch where to deploy the config
        deploy: true                      # choose from [true, false]
        profile:
          admin_state: true               # choose from [true, false]
          mode: lo                        # choose from [lo]
          int_vrf: ""                     # VRF name
          ipv4_addr: 192.169.10.1         # ipv4 address for the loopback interface
          secondary_ip: 192.168.10.2.     # secondary ipv4 address for loopback interface
          ipv6_addr: fd01::0201           # ipV6 address for the loopback interface
          route_tag: ""                   # Routing Tag for the interface
          cmds:                           # Freeform config
            - no shutdown
          description: "loopback interface 100 configurations"

```

### References

![image](https://github.com/CiscoDevNet/ansible-dcnm/assets/3794076/8a840a4f-a1cb-4be1-a8de-ff336fc07cba)

**Additional context**

PR #254, opened 2023-08-22
Add support for fabric loopback interfaces

mikewiebe:
Addresses the following issues:
  * https://github.com/CiscoDevNet/ansible-dcnm/issues/245
  * https://github.com/CiscoDevNet/ansible-dcnm/issues/246
  * Add support for mpls based fabric loopback interfaces

### DIFF
--- a/plugins/modules/dcnm_interface.py
+++ b/plugins/modules/dcnm_interface.py
@@ -1749,6 +1749,8 @@ class DcnmIntf:
             "PEER2_PO_CONF": "peer2_cmds",
             "PEER1_ACCESS_VLAN": "peer1_access_vlan",
             "PEER2_ACCESS_VLAN": "peer2_access_vlan",
+            "DCI_ROUTING_PROTO": "dci_routing_proto",
+            "DCI_ROUTING_TAG": "dci_routing_tag",
         }
 
         # New Interfaces
@@ -2804,6 +2806,16 @@ class DcnmIntf:
                 "route_tag"
             ]
 
+        # Properties for mode 'mpls' Loopback Interfaces
+        if delem[profile]["mode"] == "mpls":
+
+            # These properties are read_only properties and are not exposed as
+            # properties that can be modified.  They will be updated from the
+            # self.have dictionary to reflect the actual values later in the
+            # code workflow that walks the want values and compares to have values.
+            intf["interfaces"][0]["nvPairs"]["DCI_ROUTING_PROTO"] = "PLACE_HOLDER"
+            intf["interfaces"][0]["nvPairs"]["DCI_ROUTING_TAG"] = "PLACE_HOLDER"
+
     def dcnm_intf_get_eth_payload(self, delem, intf, profile):
 
         # Extract port id from the given name, which is of the form 'po300'
@@ -3418,6 +3430,11 @@ class DcnmIntf:
         if t_e1 != t_e2:
 
             if (state == "replaced") or (state == "overridden"):
+                # Special handling is required for mode 'mpls' loopback interfaces.
+                # They will contain either of the following two read_only properties.
+                if k == 'DCI_ROUTING_PROTO' or k == 'DCI_ROUTING_TAG':
+                    return "copy_and_add"
+
                 return "add"
             elif state == "merged":
                 # If the key is included in config, then use the value from want.
@@ -4618,15 +4635,6 @@ class DcnmIntf:
 
         path = self.paths["INTERFACE"]
         for payload in self.diff_replace:
-
-            if payload.get('policy') == 'int_mpls_loopback':
-                # Fabric mpls interfaces have two read_only properties that
-                # must be added to the payload from the self.have dictionary
-                # before attempting to modify the interface
-                drp = self.have[0]["interfaces"][0]["nvPairs"]["DCI_ROUTING_PROTO"]
-                drt = self.have[0]["interfaces"][0]["nvPairs"]["DCI_ROUTING_TAG"]
-                payload['interfaces'][0]["nvPairs"]["DCI_ROUTING_PROTO"] = drp
-                payload['interfaces'][0]["nvPairs"]["DCI_ROUTING_TAG"] = drt
 
             json_payload = json.dumps(payload)
             resp = dcnm_send(self.module, "PUT", path, json_payload)
--- a/plugins/modules/dcnm_interface.py
+++ b/plugins/modules/dcnm_interface.py
@@ -347,7 +347,7 @@ options:
           mode:
             description:
             - Interface mode
-            choices: ['lo']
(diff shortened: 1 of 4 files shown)


plugins/modules/dcnm_interface.py
```diff
@@ -3159,6 +3177,8 @@ def dcnm_get_intf_payload(self, delem, sw):
         if "aa_fex" == delem["type"]:
             self.dcnm_intf_get_aa_fex_payload(delem, intf, "profile")
 
+        # import epdb ; epdb.serve(port=8888)
```
mmudigon, line 3180: You may want to remove this

plugins/modules/dcnm_interface.py
```diff
@@ -4585,6 +4605,13 @@ def dcnm_intf_send_message_to_dcnm(self):
         path = self.paths["GLOBAL_IF"]
         for payload in self.diff_create:
 
+            # The 'int_fabric_loopback_11_1' policy is a special policy for
+            # managing fabric loopback interfaces.  These interfaces are created
+            # at fabric creation time and can only be manged by appening
+            # '/pti?isMultiEdit=false' to the path.
+            if payload.get('policy') == 'int_fabric_loopback_11_1':
+                path = path + '/pti?isMultiEdit=false'
+
```
mmudigon, line 4614: I remember we discussed regarding POST and PUT for these fabric interfaces. You mentioned that we should always use POST for these interfaces even if this interface already exist. I don't see the code for forcing POST even if a fabric interface already exist.

mikewiebe: It turns out that `PUT` does work with `/pti?isMultiEdit=false`.  I had not tested it with this setting so looks like we are ok.

tests/integration/targets/dcnm_interface/tests/dcnm/dcnm_lo_fabric.yaml
```diff
@@ -0,0 +1,199 @@
...
+      check_deploy: True
+      fabric: "{{ ansible_it_fabric }}" 
+      state: overridden                     # only choose form [merged, replaced, deleted, overridden, query]
+    register: result
+    when: IT_CONTEXT is not defined
+
+  - assert:
+      that:
+        - 'item["RETURN_CODE"] == 200'  
+    loop: '{{ result.response }}'
```
mmudigon, line 198: I assume that there will be atleast one fabric loopback interface. Otherwise the tests may not run

mikewiebe: Yes, the test will only run if Loopback0 and Loopback1 are both fabric interfaces.  I think this is a safe assumption given that we are controlling the test environment.

mikewiebe: I just made a new update to the integration tests to be more complete

dsx1123:
verified with VPC use cases, it works fine, but please add examples of modifying the loopback1 in the doc.

mikewiebe:
> verified with VPC use cases, it works fine, but please add examples of modifying the loopback1 in the doc.

Added the following example in the doc

```yaml
## Loopback Interfaces Created During Fabric Creation
- name: Mange Fabric loopback interfaces
  cisco.dcnm.dcnm_interface:
    fabric: mmudigon-fabric
    state: merged
    config:
      - name: lo1                           # This is usually lo0 or lo1 created during fabric creation
        type: lo
        switch:
          - "192.172.1.1"                   # provide the switch where to deploy the config
        deploy: true                        # choose from [true, false]
        profile:
          admin_state: false                # choose from [true, false]
          mode: fabric                      # This must be set to 'fabric' for fabric loopback interfaces
          secondary_ipv4_addr: 172.16.5.1   # secondary ipv4 address for loopback interface
          route_tag: "100"                  # Routing Tag for the interface
          cmds:                             # Freeform config
            - no shutdown
          description: "Fabric interface managed by Ansible"
```

mikewiebe:
@dsx1123 asked me to add support for another type of fabric interface so I plan to add that before we can merge this PR

mikewiebe:
**Update:** After more testing I discovered that if I am using the correct template for the loopback interfaces I don't need to add `/pti?isMultiEdit=false` to the path.  It's sufficient to just use `PUT`

plugins/modules/dcnm_interface.py
```diff
@@ -4567,6 +4619,16 @@ def dcnm_intf_send_message_to_dcnm(self):
         path = self.paths["INTERFACE"]
         for payload in self.diff_replace:
 
+            if payload.get('policy') == 'int_mpls_loopback':
```
mikewiebe, line 4622: @mmudigon if there is a better place or way to add these properties to the payload let me know what you think.  This seems fairly simple/safe but there could be a better option.

plugins/modules/dcnm_interface.py
mikewiebe: **Update:** After more testing I discovered that if the module uses the correct template for the loopback interfaces I don't need to add /pti?isMultiEdit=false to the path. It's sufficient to just use PUT

plugins/modules/dcnm_interface.py
```diff
@@ -2752,6 +2778,32 @@ def dcnm_intf_get_loopback_payload(self, delem, intf, profile):
...
+
+            intf["interfaces"][0]["nvPairs"]["SECONDARY_IP"] = delem[profile][
+                "secondary_ipv4_addr"
+            ]
+            intf["interfaces"][0]["nvPairs"]["V6IP"] = str(
+                delem[profile]["ipv6_addr"]
+            )
+            intf["interfaces"][0]["nvPairs"]["ROUTE_MAP_TAG"] = delem[profile][
+                "route_tag"
+            ]
```
mmudigon, line 2805: Assume there are no specific nvPair objects for mpls mode

tests/integration/targets/dcnm_interface/tests/dcnm/dcnm_lo_mpls.yaml
```diff
@@ -0,0 +1,151 @@
...
+
+##############################################
+##                MERGE                     ##
+##############################################
+
+    - name: Modify mpls fabric loopback interface
+      cisco.dcnm.dcnm_interface: &lo_merge
+        check_deploy: True
+        fabric: "{{ ansible_it_fabric }}"
+        state: merged
```
mmudigon, line 58: How about state replaced? Can we replace an existing fabric loopback interface with the same type but different parameters

mikewiebe:
@dsx1123 and @mmudigon This code is ready for review again.  If there are no further comments we can merge

mmudigon (approved):

PR #256, opened 2023-08-31
fix an issue when using merge to update mpls_underlay link

dsx1123:
changes in this PR:
- Fix an issue when using merge to update mpls_underlay links
- Fix a typo in the code that causes peer1_description and peer2_description are not updated correctly


mikewiebe (approved):
Reviewed with @mmudigon this morning and we are approving

PR #266, opened 2023-10-11
Fix for netcommon install issue in Git action

praveenramoorthy:
Fix for netcommon install issue in Git action CI pipeline.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -72,7 +72,7 @@ jobs:
         path: .cache/collection-tarballs
 
     - name: Install the collection tarball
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -71,6 +71,12 @@ jobs:
         name: collection
         path: .cache/collection-tarballs
 
+    - name: Create ansible.cfg file
+      run: |
+       echo "[galaxy]" > ansible.cfg
+       echo "server = https://old-galaxy.ansible.com/" >> ansible.cfg
```
mikewiebe, line 77: Interesting... so the fix is to point to the old galaxy.  Do we have any idea how long this old instance will be maintained?

PR #275, opened 2024-03-01
Fix for non zero max hop config

praveenramoorthy:
When max is set to and non zero value, and if there are other switches in the network within that hops, the discovery api returns all the switches available with that hop. In such cases inventory module tries to configure all available switches apart from the configured one. The fix now restricts the configuration only to the configured switch.

### DIFF
--- a/tests/integration/targets/dcnm_inventory/tests/dcnm/merged.yaml
+++ b/tests/integration/targets/dcnm_inventory/tests/dcnm/merged.yaml
@@ -188,6 +188,43 @@
     fabric: "{{ ansible_it_fabric }}"
     state: deleted
 
+- name: MERGED - Merge a Switch using GreenField Deployment with non zero max hop
+  cisco.dcnm.dcnm_inventory: &conf_hop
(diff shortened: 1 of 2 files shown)


tests/integration/targets/dcnm_inventory/tests/dcnm/merged.yaml
```diff
@@ -188,6 +188,43 @@
...
+- name: MERGED - Merge a Switch using GreenField Deployment with non zero max hop
+  cisco.dcnm.dcnm_inventory: &conf_hop
+    fabric: "{{ ansible_it_fabric }}"
+    state: merged
+    config:
+    - seed_ip: "{{ ansible_switch1 }}"
+      auth_proto: MD5
+      user_name: "{{ switch_username }}"
+      password: "{{ switch_password }}"
+      max_hops: 2
```
mikewiebe, line 200: Update documentation to indicate this will always be set internally to 0 and update the code to set this to 0 regardless of what the user passes in.

mikewiebe (approved):

Issue #278, opened 2024-03-22
dcnm_interface expecting case sensitive speed value "Auto" for Routed Interfaces


rrahimm:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Ansible Version and collection version

ansible 2.13.9 1e80e4319f

collections:

name: ansible.utils
version: 2.11.0
name: ansible.netcommon
version: 5.3.0
name: cisco.nxos
version: 5.3.0
name: https://github.com/CiscoDevNet/ansible-dcnm.git
version: main

### DCNM version

V 12.1.3b

### Affected module(s)

dcnm_interface

### Ansible Playbook

```yaml
- name: Manage Interface Routed
  cisco.dcnm.dcnm_interface:
    fabric: "{{ MD.fabric.name }}"
    state: replaced
    config:
      - name : Ethernet1/11
  		type: eth
  		switch: 
  		  - 192.168.123.59
  		deploy: false
  		profile:
  		  admin_state: True
  		  mode: 'routed'
  		  ipv4_addr: 192.168.57.1
  		  ipv4_mask_len: 24
  		  ipv6_addr: 2021:10:1:1::1
  		  ipv6_mask_len: 64
  		  description: topology_switch_routed_interface
  		  route_tag: 12345
  		  int_vrf: NetAsCodeVrf1
  		  mtu: 1500
  		  speed: auto
 ```

### Debug Output

        "DATA": [
            {
                "column": 0,
                "entity": "netascode-rtp-spine1:Ethernet1/8",
                "line": 0,
                "message": "Validation failed for following fields: [SPEED]",
                "reportItemType": "ERROR"
            }
        ],
        "MESSAGE": "Internal Server Error",
        "METHOD": "PUT",
        "REQUEST_PATH": "https://rtp-ndfc1.cisco.com:443/appcenter/cisco/ndfc/api/v1/lan-fabric/rest/interface",
        "RETURN_CODE": 500
    }

### Expected Behavior

speed value in the playbook is case-insensitive for other profile types.
However expects "Auto" case-sensitive for type >> mode: 'routed'

[...]

Issue #276, opened 2024-03-13
Unable to change Interface speed using dcnm_interface


rrahimm:
### Issue Description:

Unable to change Interface speed using dcnm_interface.

Unable to change Interface speed using dcnm_interface when the only parameter being changed is Speed.
But if we attempt to change one more parameter along with speed, then both parameters get updated.

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Ansible Version and collection version

ansible 2.13.9 1e80e4319f

collections:
  - name: ansible.utils
    version: 2.11.0
  - name: ansible.netcommon
    version: 5.3.0
  - name: cisco.nxos
    version: 5.3.0
  - name: https://github.com/CiscoDevNet/ansible-dcnm.git
     version: main

### DCNM version

* V 12.1.3b

### Affected module(s)

* dcnm_interface

### Ansible Playbook

```yaml
- name: Manage Interface Routed
  cisco.dcnm.dcnm_interface:
    fabric: "{{ MD.fabric.name }}"
    state: replaced
    config:
      - name : Ethernet1/11
  		type: eth
  		switch: 
  		  - 192.168.123.59
  		deploy: false
  		profile:
  		  admin_state: True
  		  mode: 'routed'
  		  ipv4_addr: 192.168.57.1
  		  ipv4_mask_len: 24
  		  ipv6_addr: 2021:10:1:1::1
  		  ipv6_mask_len: 64
  		  description: topology_switch_routed_interface
  		  route_tag: 12345
  		  int_vrf: NetAsCodeVrf1
  		  mtu: 1500
  		  speed: 1Gb
```

### Debug Output

```yaml
TASK [cisco.nac_dc_vxlan.create : Manage Interface Routed] ***************************************************************
task path: /interfaces.yml:51
ok: [netascode_fabric2] => {
    "changed": false,
    "diff": [
        {
            "debugs": [],
            "deferred": [],
            "delete_deploy": [],
            "deleted": [],
            "deploy": [],
            "merged": [],
            "overridden": [],
            "query": [],
            "replaced": [],
            "skipped": []
        }
    ],
[...]

PR #281, opened 2024-03-29
Fix for interface module, Issues 276 and 278

mmudigon:
This commit includes the fixes for issues #276 and #278
Speed parameter was not processed in the current implementation and thus idempotence test cases w,r,t speed failed.
Code is now updated to process speed parameter and update accordingly.
This commit also includes updates to accept speed parameter through playbook in a case insensitive way. Code is now updated to handle these cases.


mikewiebe:
Fix tested in the context of our services as code project and looks to be working.

mikewiebe (approved):

PR #287, opened 2024-04-23
Code update to handle ethernet interface delete/default based on switch role

mmudigon:
DCNM interface module has been updated to check for switch role and default ethernet interfaces to appropriate policy. for e.g. a switch with role as 'leaf' will be defaulted to policy trunk and switches with other roles will be defaulted to routed policy.


mikewiebe:
Tested this as part of the SaC solution and it is working as expected

mikewiebe (approved):

PR #290, opened 2024-04-27
Update dcnm_image_upgrade action plugin timeout to 1800 seconds.

allenrobel:
This is to update the timeout value in the ActionModule for dcnm_image_upgrade to 1800 seconds.


dsx1123:
LTGM

PR #298, opened 2024-05-29
Fix to override only specified interface types instead of all types of interfaces

mmudigon:
By default interface module will delete/default all interface types during overridden/deleted states. The module has been updated to do the following:

- provide an optional list parameter in the playbook to include specific interface types
- if this list is empty, then the behaviour will be similar to what is already implemented i.e. all interface types will be handled during overridden and deleted states
- if this list includes specific interface types, then only those interface types will be handled during overridden and deleted states.
- this commit also fixes a couple of deploy issues during overridden/deleted states.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,8 +14,6 @@ This collection is intended for use with the following release versions:
 
 This collection has been tested against following Ansible versions: **>=2.9.10**.
 
-For collections that support Ansible 2.9, please ensure you update your `network_os` to use the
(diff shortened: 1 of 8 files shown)


mikewiebe:
@mmudigon Thanks for opening this PR.  Looking forward to reviewing/testing it when I get back from Cisco Live

mikewiebe (approved):

PR #300, opened 2024-06-06
Case agnostic checks for interface names in dcnm_links module

mmudigon:
Interface names can be specific to type of box used. In some cases a loopback interface may be like loopback100 and in some cases it will be like Loopback100. To make this transparent to the user, dcnm_links module has been updated to handle these cases.

### DIFF
--- a/tests/unit/modules/dcnm/fixtures/dcnm_links_payloads.json
+++ b/tests/unit/modules/dcnm/fixtures/dcnm_links_payloads.json
@@ -328,7 +328,7 @@
         "link-uuid": "LINK-UUID-1446600",
         "sw2-info": {
           "fabric-name": "mmudigon-numbered",
-          "if-name": "Ethernet1/1",
+          "if-name": "ethernet1/1",
           "sw-sys-name": "n9kv-num-2",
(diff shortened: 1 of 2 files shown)


tests/unit/modules/dcnm/fixtures/dcnm_links_payloads.json
```diff
@@ -328,7 +328,7 @@
         "link-uuid": "LINK-UUID-1446600",
         "sw2-info": {
           "fabric-name": "mmudigon-numbered",
-          "if-name": "Ethernet1/1",
+          "if-name": "ethernet1/1",
```
mikewiebe, line 331: This is the mock response coming back from NDFC right?  Does it come back as upper case or lower case?

mmudigon: Yes. But whether it is upper or lower depends on the switch model. Code is now updated to handle both cases appropriately. To check that it does, I have changes existing test cases to have both upper and lower cases for interface names

mikewiebe (approved):

PR #335, opened 2024-10-14
VerifyPlaybookParams(): parameter validation fixes

allenrobel:
# Fixes for incorrect parameter validation

1. VerifyPlaybookParams().update_decision_set()

Conditional for ``playbook_is_valid`` should compare to False explicitly since ``playbook_is_valid`` might be None.

2. VerifyPlaybookParams().update_decision_set() 

If checks for all sources for the parameter value (playbook, controller, template default) return None (i.e. the parameter should not be considered when validating the dependent parameter), then we now set the decision_set to {True} (meaning "accept" the parameter as valid).

3. Add a knob ``skip_validation`` which, if True, will send the playbook to NDFC without verifying its content.  The idea being that this knob will help determine if parameter validation is disallowing something that NDFC would accept.  Default for ``skip_validation`` is False.

4. Improve ``KeyError()`` message content to include class_name and method_name.

5. plugins/modules/dcnm_fabric.py: EXAMPLE section. Added example using ``skip_validation``

6. plugins/modules/dcnm_fabric.py: EXAMPLE section. Added example that generates an error message and provides interpretation of the error message.


.github/workflows/main.yml
```diff
@@ -35,13 +35,13 @@ jobs:
       run: pip install https://github.com/ansible/ansible/archive/v${{ matrix.ansible }}.tar.gz --disable-pip-version-check
 
     - name: Build a DCNM collection tarball
-      run: ansible-galaxy collection build --output-path "${GITHUB_WORKSPACE}/.cache/collection-tarballs"
+      run: ansible-galaxy collection build --output-path "${GITHUB_WORKSPACE}/.cache/v${{ matrix.ansible }}/collection-tarballs"
```
mikewiebe, line 38: Are these changes you made or do you need to merge `develop` into this branch?

plugins/modules/dcnm_fabric.py
```diff
@@ -37,6 +37,11 @@
         description:
         - The state of the feature or object after module completion
         type: str
+    skip_validation:
+        default: false
+        description:
+        - Skip playbook parameter validation.  Useful for debugging.
```
mikewiebe, line 43: If the user enables this flag but still passes invalid keys/values to NDFC, then we are relying on NDFC to return a meaningful error message in that case right?

.github/workflows/main.yml
allenrobel: I did pull the changes fron develop into this branch.  These must have been committed to develop before I pulled it.  No, I didn't make these changes.

plugins/modules/dcnm_fabric.py
allenrobel: Yes, that's correct.

mikewiebe (approved):

PR #336, opened 2024-10-25
Relative imports for shared code

allenrobel:
When other projects try to leverage the code in ansible-dcnm, it results in import errors if ansible-dcnm is not installed under ansible_collections/cisco/dcnm.

This PR changes the imports for all code that is shareable such that it can be imported wherever the ansible-dcnm  repository happens to be cloned.

This PR also includes some new endpoints and endpoint parsers for Federations and OneManage, as well as  sender_requests.py.  None of these are used by any modules currently.

Code modified includes only that which is possible to be shared by other projects (i.e. code that is not dependent on Ansible).

module_utils/common/api/*
module_utils/common/controller_features.py
module_utils/common/controller_version.py
module_utils/common/ep/*
module_utils/common/epp/*
module_utils/common/image_policies.py
module_utils/common/maintenance_mode.py
module_utils/common/maintenance_mode_info.py
module_utils/common/rest_send.py
module_utils/common/rest_send_v2.py
module_utils/common/sender_dcnm.py
module_utils/common/sender_requests.py
module_utils/common/switch_details.py
module_utils/fabric/*
module_utils/image_policy/*
module_utils/image_upgrade/*


plugins/module_utils/common/ep/nexus/api/api.py
```diff
@@ -0,0 +1,43 @@
...
+
+class Api(Nexus):
+    """
+    ## Nexus Dashboard enpoints - Nexus().Api()
+
+    ### Description
+    Common methods and properties for ND subclasses.
+
+    ### Path
+    ``/nexus/api``
```
mikewiebe, line 33: Do you know why the root of this endpoint name is `/nexus/`?

mikewiebe (approved):
@allenrobel LGTM.  Just had one question in the comments

plugins/module_utils/common/ep/nexus/api/api.py
allenrobel: This is the root for Nexus Dashboard endpoints, which is used by EppFederationMemberByName().  Though, this class is not yet being used (I figured it would be needed later for the OneManage stuff).

If you'd prefer, I could remove the Federation things for now and add them only if/when needed for OneManage.

mikewiebe: No need to remove.  Just curious

PR #340, opened 2024-11-08
Release 3.6.0

mikewiebe:


allenrobel (approved):
Generally looks good.  I took some notes about various typos in docstrings that we can circle back later to fix.

PR #345, opened 2024-11-20
Module: dcnm_vpc_pair - Remove Manageable Check for Devices

mikewiebe:
The `dcnm_vpc_pair` module currently has a check to see if devices are in a `manageable` state before attempting to put them into a vpc pair but this check is not needed and in fact breaks pre-provision workflows.

A device or switch state can be unmanageable for a variety of reasons but we should not be checking this state to use it for making configuration decisions.  A device that is in this state can still be configured as a vpc pair.

Removing this check but leaving the function in the library if we decide we need it down the road for different use cases.

### DIFF
--- a/plugins/modules/dcnm_vpc_pair.py
+++ b/plugins/modules/dcnm_vpc_pair.py
@@ -436,7 +436,6 @@ from ansible_collections.cisco.dcnm.plugins.module_utils.network.dcnm.dcnm_vpc_p
     dcnm_vpc_pair_utils_get_sync_status,
     dcnm_vpc_pair_utils_get_delete_list,
     dcnm_vpc_pair_utils_get_all_filtered_vpc_pair_pairs,
-    dcnm_vpc_pair_utils_validate_devices,
 )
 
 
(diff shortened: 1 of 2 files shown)


plugins/modules/dcnm_vpc_pair.py
```diff
@@ -1104,9 +1103,6 @@ def dcnm_vpc_pair_translate_playbook_info(self, config, ip_sn, hn_sn):
             # Add other translations as required
             dcnm_vpc_pair_utils_translate_config(self, cfg)
 
-            # Check if the switches included in the config are Manageable.
-            dcnm_vpc_pair_utils_validate_devices(self, cfg)
```
allenrobel, line 1108: This is the only place in the DCNM collection that dcnm_vpc_pair_utils_validate_devices() is used.

If we don't anticipate using this function in the future, was wondering if we should also remove dcnm_vpc_pair_utils_validate_devices() from module_utils/network/dcnm/dcnm_vpc_pair_utils.py?

And if dcnm_vpc_pair_utils_validate_devices() is removed, then dcnm_vpc_pair_utils_check_if_meta() in the same file can also be removed since it's accessed only from dcnm_vpc_pair_utils_validate_devices()

allenrobel (approved):
Approved with one comment.

plugins/modules/dcnm_vpc_pair.py
mikewiebe: Yes I agree, I will remove it

PR #346, opened 2024-11-20
Module: dcnm_interface - Remove Manageable Check for Interfaces

mikewiebe:
The `dcnm_interface` module currently has a check to see if devices are in a `manageable` state before attempting to create interfaces but this check is not needed and in fact breaks pre-provision workflows.

A device or switch state can be unmanageable for a variety of reasons but we should not be checking this state to use it for making configuration decisions.


allenrobel (approved):
LGTM

Issue #368, opened 2025-02-03
Inventory addition to fabric fails when default creds are not set


pritispa:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Ansible Version and collection version

Ansible core: 2.18.0
dcnm collection: 3.5.0

### DCNM version

* ND 3.2.1i/NDFC 12.2.2

### Affected module(s)

* dcnm_inventory

### Ansible Playbook

Task:
```yaml
- name: Add Switches to VXLAN Fabric
  cisco.dcnm.dcnm_inventory:
    fabric: "{{ vxlan_fabric_parameters.FABRIC_NAME }}"
    state: merged
    config: "{{ vxlan_switch_list }}"
  register: result
```
Vars:
```
vxlan_switch_list:
  - seed_ip: [IP_REMOVED]
    auth_proto: MD5
    max_hops: 0
    role: spine
    preserve_config: false
    user_name: admin
    password: "pwdremoved"
  - seed_ip: [IP_REMOVED]
    auth_proto: MD5
    max_hops: 0
    role: spine
    preserve_config: false
    user_name: admin
    password: "pwdremoved"
  - seed_ip: [IP_REMOVED]
    auth_proto: MD5
    max_hops: 0
    role: leaf
    preserve_config: false
    user_name: admin
    password: "pwdremoved"
  - seed_ip: [IP_REMOVED]
    auth_proto: MD5
    max_hops: 0
    role: leaf
    preserve_config: false
    user_name: admin
    password: "pwdremoved"
  - seed_ip: [IP_REMOVED]
    auth_proto: MD5
    max_hops: 0
    role: leaf
    preserve_config: false
    user_name: admin
    password: "pwdremoved"
  - seed_ip: [IP_REMOVED]
    auth_proto: MD5
    max_hops: 0
    role: leaf
    preserve_config: false
    user_name: admin
    password: "pwdremoved"
```

### Debug Output

https://gist.github.com/pritispa/3cd01036e1d62493fcaa802d2b86d502
### Expected Behavior
Without Default creds being set in NDFC, there are two options to handle this situation:
[...]

PR #370, opened 2025-02-05
Issue 368: #368

prabahal:
#368 

**Testing Completed for Older NDFC version**
Issue:
Switch addition to inventory fails if Default credentials are not set in NDFC.

Solution:

- While discovering the switch adding the additional attribute discoveryCredForLan to discover using individual switch credentials.
- Also added dcnm_fabric IT issue fix as well.


mikewiebe:
@prabahal Tested this fix and it looks to be working.  Thanks!

mikewiebe (approved):

Issue #372, opened 2025-02-12
Interface module resets PO and VPC members (previous module behaviour used to check if member ports were part of PO and throw error)


mthurstocisco:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### DCNM version

* V 3.6.0

### Affected module(s)

* dcnm_interface

### Ansible Playbook

```yaml
# Copy-paste your anisble playbook here 
```

### Expected Behavior

When trying to configure a trunk or access interface that is part of a port channel or vpc, module should not allow update of most aspects of the interface (as per the GUI, if the policy on the port is int_vpc_trunk_po_member_11_1 then the only editable fields are Interface Description, Freeform Config and Interface Admin State). When trying to update other fields the module should throw an error to say the port is part of a port channel or VPC (as per older versions of the module).

### Actual Behavior

The member interface gets completely reset, all the po/vpc membership config is removed from the port thus removing the interface from the po/vpc.

### Steps to Reproduce

Configure a VPC or PO, add a member interface then try to update any aspect of the member interface and the interface will be removed from the VPC/PO. In the GUI the VPC/PO will still show the member ports in its configuration but all channel membership configuration will be removed from the member interface.

### References

Having had a quick look at the code, it appears that anytime there is a difference between the policy on a port as part of the input to the playbook and the policy on the interface on NDFC then no key/value comparison is done and the input to the playbook is just pushed as a whole to NDFC. 

As I read it, there will need to be some special cases in the key/value comparison to handle po/vpc membership and to make sure we push the correct (existing in NDFC) policy back to the interface in NDFC

mthurstocisco:
@mikewiebe

PR #382, opened 2025-03-04
Support for modification of PortChannel and vPC Member Interfaces

mikewiebe:
**Summary:**

Resolves https://github.com/CiscoDevNet/ansible-dcnm/issues/372

Adds integration tests for the issue but uses a new approach
  * Adds new directory for the test playbooks under `playbooks/roles/ndfc_interface` to keep the new style tests separate from the old style tests.
  * Uses Jinja2 templates to generate the module config payloads required from the test data. The jinja2 files generate test files under `tests/integration/targets/ndfc_interface/files/` which are then read into Ansible variables and passed to the modules.
  * Uses an action plugin (plugins/action/tests/unit/ndfc_pc_members_validate.py) to validate data that comes back from an NDFC query against the user supplied test data
  * This test does not set the deploy flag to true in an effort to speed up the test.


plugins/action/tests/unit/ndfc_pc_members_validate.py
```diff
@@ -0,0 +1,120 @@
...
+            return results
+
+        # ReWrite List Data to Dict keyed by interface name
+        ndfc_data_dict = {}
+        for interface in ndfc_data['response']:
+            int = interface['interfaces'][0]['ifName']
+            ndfc_data_dict[int] = interface['interfaces'][0]
+            ndfc_data_dict[int]['policy'] = interface['policy']
+
+        for int in interface_list:
```
allenrobel, line 56: int is a reserved Python keyword, but as long as we're not using it specifically e.g. int("10") then shouldn't be a problem.

plugins/modules/dcnm_interface.py
```diff
@@ -419,6 +419,9 @@
           mode:
             description:
             - Interface mode
+            - When ethernet interface is a PortChannel or vPC member mode is ignored and
+              the only properties that can be managed are 'admin_state', 'description' and 'cmds'.
+              All other properties are ignored.
```
allenrobel, line 424: Some awkward wording here.  Should it be like below (removing "and" from "ignored and" and replacing with "ignored, "

When ethernet interface is a PortChannel or vPC member, mode is ignored.  The only properties that can be managed for PortChannel or vPC member interfaces are 'admin_state', 'description' and 'cmds'.   All other properties are ignored.

allenrobel (approved):
LGTM, approving with just a couple minor comments.

plugins/action/tests/unit/ndfc_pc_members_validate.py
mikewiebe: Thanks Allen, I changed it from `int` to `interface`

plugins/modules/dcnm_interface.py
mikewiebe: Updated

PR #391, opened 2025-03-29
Issue #352 : more preparatory changes

allenrobel:
# Summary

This PR for `dcnm_vrf` includes the following:

1. Modifies method `send_to_controller()` to add a parameter `log_response` that controls whether the controller response is added to the Ansible playbook result output.  This allows us to suppress results that would not be of interest to the user.
2. Simplifies a conditional in `get_have()`
3. Fixes a pylint error in `release_resources_by_id()`
4. Cleans up the merged integration test
5. Adds an `__init__.py` file to appease pylint

## Detail

1. plugins/modules/dcnm.vrf

- send_to_controller()

Add a log_response parameter which controls whether the controller response will be added to the ansible response.

- get_have() - simplify conditional

attach_state = False if attach["lanAttachState"] == "NA" else True

Changed to:

attach_state = not attach["isLanAttached"] == "NA"

- release_resources_by_id()

Fix pylint error (id_list was being redefined in the for loop, so renamed to item).

- failure(): disable pylint check for protected-access

2. tests/integration/targets/dcnm_vrf/tests/dcnm/merged.yaml

There are no functional changes here.  We've only done the following:

- Remove single-quotes from all asserts
- Remove jinja2 templating around variables in asserts (these generate warnings and are not needed).
- Removing the single-quotes broke asserts of the following form:

- "test phrase" in var

To fix this, added set fact for "test phrase" like so:

```yaml
- name: set fact
  set_fact: TEST_PHRASE: "test phrase"

- assert:
  that:
    - TEST_PHRASE in var
```

6. plugins/__init__.py

- Added to appease pylint

### DIFF
--- a/plugins/modules/dcnm_vrf.py
+++ b/plugins/modules/dcnm_vrf.py
@@ -595,6 +595,11 @@ dcnm_vrf_paths = {
 
 
 class DcnmVrf:
+    """
+    # Summary
+
+    dcnm_vrf module implementation.
+    """
 
     def __init__(self, module):
         self.class_name = self.__class__.__name__
@@ -1641,7 +1646,7 @@ class DcnmVrf:
             attach_list = vrf_attach["lanAttachList"]
             deploy_vrf = ""
             for attach in attach_list:
-                attach_state = False if attach["lanAttachState"] == "NA" else True
+                attach_state = not attach["isLanAttached"] == "NA"
                 deploy = attach["isLanAttached"]
                 deployed = False
                 if deploy and (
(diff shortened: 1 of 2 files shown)


mikewiebe (approved):
LGTM

plugins/modules/dcnm_vrf.py
```diff
@@ -2874,7 +2898,12 @@ def push_diff_delete(self, is_rollback=False):
                 continue
             delete_path = f"{path}/{vrf}"
             self.send_to_controller(
-                action, verb, delete_path, self.diff_detach, is_rollback
+                action,
+                verb,
+                delete_path,
+                self.diff_delete,
```
mikewiebe, line 2904: Approved these changes but at some point we need to test this change with our VXLAN as Code solution to make sure we don't introduce any issues with this change.

allenrobel: Yep, agree!  I did run the integration tests, which worked.  But they also worked before this change, which is odd. 🤷‍♂️

PR #396, opened 2025-04-12
dcnm_network: Fix for issue 395

allenrobel:
# SUMMARY

This fixes an issue when networks are created in an MSD fabric, the following error happens.

```
“Invalid : Fabric mode is not multicast and Multicast Address: 239.1.1.1 is present”
```

The actual diffs for this commit are limited to items 1b and 1c below.  The rest of changes you can ignore since they are just the black linter reformatting things.

## Tests

- Ran all `dcnm_network` integration tests (all are passing)
- Ran `dcnm_network` unit tests (all are passing except two that are no longer valid)

## Changes

1. plugins/modules/dcnm_network.py

- 1a. Run through linters (sorry, this resulted in a large diff)
- 1b. get_fabric_multicast_group_address() - new method
- 1c. validate_input() - leverage get_fabric_multicast_group_address()

Two more commits will be added as part of this PR

### Commit for Unit test updates

2. tests/unit/modules/dcnm/fixtures/dcnm_network.json

- Ran through a JSON prettier.  There are no other changes to this file, so don't spend time reviewing it.

3. test/unit/modules/dcnm/test_dcnm_network.py

- Commented out two test cases that are now failing.  These are no longer valid cases though.

### Commit for integration test updates

- Removed most single-quotes from around the asserts (a couple are still needed).
- Sorted the asserts for better readability and easier maintenance.
- Changed the interface name vars as follows
   - ansible_sw1_int1 -> interface_1a
   - ansible_sw1_int2 -> interface_1b
   - ansible_sw1_int5 -> interface_1c
   - ansible_sw1_int6-> interface_1d
   - ansible_sw2_int1 -> interface_2a
   - ansible_sw2_int2 -> interface_2b
   - ansible_sw2_int5 -> interface_2c
   - ansible_sw2_int6 -> interface_2d


plugins/modules/dcnm_network.py
```diff
@@ -634,25 +628,42 @@ def diff_for_attach_deploy(self, want_a, have_a, replace=False):
                                                         else []
                                                     )
 
-                                                    if sorted(h_tor_ports) != sorted(w_tor_ports):
+                                                    if sorted(h_tor_ports) != sorted(
```
mikewiebe, line 631: Not sure why black is reformatting things like this.  I believe black was run against these files previously but I guess maybe the settings were different?  For this change and changes like it I actually prefer the single line vs the multiline reformat.  Makes it a bit harder to read IMO.

mikewiebe: What I typically do is turn off the black formatter in vscode for things like this so I don't end up with a much larger set of diffs when what I am changin

plugins/modules/dcnm_network.py
```diff
@@ -2680,22 +2769,50 @@ def push_to_remote(self, is_rollback=False):
...
+        Call fail_json for any unhandled combinations of REPLICATION_MODE and
+        version.
+        """
+        controller_version: int = 12
+        if self.dcnm_version is not None:
+            controller_version = self.dcnm_version
+
+        fabric_multicast_group_address: str = ""
+        replication_mode: str = self.fabric_det.get("nvPairs", {}).get(
+            "REPLICATION_MODE", "Ingress"
```
mikewiebe, line 2793: Might be a good ideal to make `Ingress` and `Multicast` strings lower case for comparison.  Unlikely that the controller would change this but possibly safer.

tests/unit/modules/dcnm/test_dcnm_network.py
```diff
@@ -625,13 +625,14 @@ def test_dcnm_net_replace_with_no_atch(self):
...
         self.assertEqual(result["response"][1]["RETURN_CODE"], self.SUCCESS_RETURN_CODE)
 
-    def test_dcnm_net_replace_without_changes(self):
-        set_module_args(
-            dict(state="replaced", fabric="test_network", config=self.playbook_config)
-        )
-        result = self.execute_module(changed=False, failed=False)
-        self.assertFalse(result.get("diff"))
-        self.assertFalse(result.get("response"))
+    # def test_dcnm_net_replace_without_changes(self):
```
mikewiebe, line 628: Can you add comments indicating why these are no longer valid or are you recommending we simply remove these?

plugins/modules/dcnm_network.py
allenrobel: Totally agree about black Mike.  The problem for me is that (and maybe this was only when I first started submitting PRs) black used to cause Github sanity failures if I didn't run it locally.

I think the solution for me is to review the black (and isort) configs for the ansible-dcnm repo and make sure my local settings are identical.  I AM using the requirements.txt and test-requirements.txt files to install things, so hopefully am using the same versions of black/isort.

mikewiebe:
Since you updated the integration test variable names, can you also update the variable names [here](https://github.com/CiscoDevNet/ansible-dcnm/tree/develop/playbooks/roles/dcnm_network)?

plugins/modules/dcnm_network.py
mikewiebe: Odd that you saw sanity failures.  I don't run black by default so most of my PRs lately only had what I explicitly changed so we will need to investigate that if that happens again.  For this PR it's fine since the black changes and fix changes are fairly obvious.

tests/unit/modules/dcnm/test_dcnm_network.py
allenrobel: Sure, Mike, I'll add a comment here.  These aren't valid because they DO result in changes now, and trying to modify the test cases, or the code so that there are NO changes just causes many other unit tests to fail (because of the interdependence of all the unit tests due to self.execute_module() and strict order of side effects).

Since IT passes with no modifications to the playbook asserts, I figured it would be OK to comment these out.  But I didn't delete them outright in case we want to revisit in the future.

allenrobel: Added comments to both tests.

plugins/modules/dcnm_network.py
allenrobel: Done.

mikewiebe (approved):

PR #415, opened 2025-05-15
add route_tag for subinterface

ccoueffe:
add route_tag under subinterface.


devegupt:
@ccoueffe @mikewiebe I was testing this and now I can see route tag create/update working for the sub-routed interface.

tests/unit/modules/dcnm/fixtures/dcnm_intf_multi_intf_configs.json
```diff
@@ -130,7 +130,7 @@
           "no shutdown"
         ],
         "peer1_allowed_vlans": "none",
-	"peer1_native_vlan": "10",
+	      "peer1_native_vlan": "10",
```
mikewiebe, line 133: Looks like white space was introduced here and elsewhere.. was that intentional?  Can you restore the original back?

mikewiebe (approved):

Issue #80, opened 2021-07-10
dcnm_network requires ports attribute when attaching to switch


gve-vse-tim:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Ansible Version and collection version

Ansible 2.10.5
DCNM Collection 1.1.1

### DCNM version

* V 11.5.1

### Affected module(s)

* dcnm_network

### Ansible Playbook

```yaml
    - name: Create a Network and Attach to Switches
      dcnm_network:
        fabric: multisite_domain_fabric
        state: replaced
        config:
          - net_name: VLAN10
            vrf_name: VXLAN_OVERLAY
            vlan_id: 10
            gw_ip_subnet: '192.168.10.1/24'
            net_id: 30000
            deploy: true
            attach:
              - ip_address: 10.60.66.131

```

### Debug Output

```
TASK [Create a Network and Attach to Switches]
*******************************************************************
fatal: [dcpod_dcnm]: FAILED! => {"changed": false, "msg": "Invalid parameters in playbook: ports : Required parameter not found"}
```

### Expected Behavior

The VLAN (and SVI) should be configured on the switch specified without any errors and without the requirement to assign it to a port.

### Actual Behavior

The DCNM module has specified that ports is a mandatory attribute for attaching the network to a switch.  This behavior does not match the DCNM GUI behavior.

It's also a case of DRY (don't repeat yourself) in that you have to define allowed VLANs on trunk or access VLANs when configuring portchannel,  VPC, or other trunk interfaces.  So, customers (and myself) prefer defining VLAN members on the actual interfaces and not on the networks.

praveenramoorthy:
@gve-vse-tim Thanks for raising this SR. Port is a mandatory attribute, but you can have an empty list for that attribute.

    - name: Create a Network and Attach to Switches
      dcnm_network:
        fabric: multisite_domain_fabric
        state: replaced
        config:
          - net_name: VLAN10
            vrf_name: VXLAN_OVERLAY
            vlan_id: 10
            gw_ip_subnet: '192.168.10.1/24'
            net_id: 30000
            deploy: true
[...]

PR #427, opened 2025-06-09
Fix for Issue #80 "dcnm_network requires ports attribute when attaching to switch"

AKDRG:
Issue:
https://github.com/CiscoDevNet/ansible-dcnm/issues/80

Code Changes Description:
From the ports parameter in the DCNM Network Class, remove "required" constraint and add default value.


mikewiebe (approved):
Looks good @AKDRG.  Thanks for fixing this!

I also updated the solution collection to handle this

https://github.com/netascode/ansible-dc-vxlan/pull/420

Issue #428, opened 2025-06-11
dcnm_fabric module returns incorrect result when querying non-existent or invalid fabrics in the NDFC


AKDRG:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Ansible Version and collection version

ansible [core 2.18.5]
collection version 3.8.0

### DCNM version

* V 12.2.2

### Affected module(s)

* dcnm_fabric

### Ansible Playbook

```yaml
# Sample Play
  - name: Query fabric
    cisco.dcnm.dcnm_fabric:
      state: query
      config:
      - FABRIC_NAME: Fabric-Test
    register: query_result
```

### Debug Output
 
ok: [192.168.1.2] => {
{
    "result": {
        "changed": false,
        "diff": [
            {
                "sequence_number": 1
            }
        ],
        "failed": false,
        "metadata": [
            {
                "action": "fabric_query",
                "check_mode": false,
                "sequence_number": 1,
                "state": "query"
            }
        ],
        "response": [
            {
                "DATA": [],
                "MESSAGE": "OK",
                "METHOD": "GET",
                "REQUEST_PATH": "https://192.168.1.2:443/appcenter/cisco/ndfc/api/v1/lan-fabric/rest/control/fabrics",
                "RETURN_CODE": 200,
                "sequence_number": 1
            }
        ],
        "result": [
            {
                "found": true,
                "sequence_number": 1,
                "success": true
            }
        ]
    }
}

### Expected Behavior

Since the queried fabric name is not present in the NDFC, "found" and "success" fields in inner "result" should be "false".

### Actual Behavior

"Found" and "success" fields in the inner "result" has the value "true".

### Steps to Reproduce

1. Using dcnm_fabric, query an NDFC host with an invalid fabric name.

mikewiebe:
Resolved by https://github.com/CiscoDevNet/ansible-dcnm/pull/429

PR #429, opened 2025-06-11
Fix for Issue #428 "dcnm_fabric returns incorrect results when querying invalid fabrics on NDFC"

AKDRG:
Issue:
https://github.com/CiscoDevNet/ansible-dcnm/issues/428

Code Changes Description:
1) With the response obtained from the NDFC, based on the fabric(s) to be queried, the add_to_diff dictionary is populated. This dictionary will contain the data of the fabrics under the "Query" operation. 
2) If add_to_diff is empty, then not one of the queried fabrics is available in the NDFC. Then, the inner result fields "found" and "success" are modified to "False" and then the final results is built.


AKDRG:
Hi Allen,
Thank you! I've fixed the UT Cases and now all tests have passed.

mikewiebe:
LGTM.  The approval link is not working for some reason (GitHub issue) but will merge.

Issue #337, opened 2024-10-30
add "Enable L3VNI w/o VLAN" support on `dcnm_vrfs` module


dsx1123:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description
There is a new way to create l3vni without vlan, it needs to be enabled per fabric setting and can also be overridden on the vrf level, the request is to support the overridden option when create the vrf, the nob is called "Enable L3VNI w/o VLAN"

suggested parameter name: `l3vni_wo_vlan`

### New or Affected modules(s):

* dcnm_vrfs

### DCNM version

* V 12.1.3b

### Potential ansible task config

```yaml
# Copy-paste your ansible playbook
```

### References

**Additional context**

mikewiebe:
Resolved by https://github.com/CiscoDevNet/ansible-dcnm/pull/435 and https://github.com/CiscoDevNet/ansible-dcnm/pull/481

PR #435, opened 2025-06-17
dcnm_vrf: Add L3VNI w/o VLAN support, #337

AKDRG:
Adding L3VNI w/o VLAN option support for dcnm_vrf module, requested through #337. 

Using "l3vni_wo_vlan" as the config parameter, which takes in "true" or "false" from the user end, to enable or disable the feature. By default, it is disabled (false).
JSON VRF Template Payload parameter, "enableL3VniNoVlan" is used to pass on the information to the NDFC.


AKDRG:
Please note that NDFC auto-handles the incompatible parameters, when the feature is toggled and doesn't push those configs to the Switches.

Issue #441, opened 2025-06-24
RP_COUNT Integer Validation Fails - Type Conversion Inconsistency in Parameter Rules


skaszlik:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Ansible Version and collection version

ansible [core 2.16.3]
python version = 3.10.14 
jinja version = 3.1.6

### DCNM version

NDFC "version": "12.2.2.241"

### Affected module(s)

ansible_collections/cisco/dcnm/plugins/module_utils/fabric/verify_playbook_params.py

### Ansible Playbook

```yaml
---
- hosts: nac-fabric1
  any_errors_fatal: true
  collections:
    - cisco.dcnm

  vars:
    fabric_name: "nac-fabric1" 
    switch_ip: "198.18.133.14"

  tasks:    
    - name: Update fabrics (nac test)
      cisco.dcnm.dcnm_fabric:
        state: merged
        config:
        - FABRIC_NAME: nac-fabric1
          FABRIC_TYPE: VXLAN_EVPN
          DEPLOY: False

          BGP_AS: "65011"
          UNDERLAY_IS_V6: False
          FABRIC_INTERFACE_TYPE: p2p
          SUBNET_TARGET_MASK: 30
          LINK_STATE_ROUTING: ospf
          ANYCAST_GW_MAC: 12:34:56:78:90:00

          REPLICATION_MODE: Multicast
          RP_COUNT: 4
          RP_MODE: bidir
          RP_LB_ID: 254
          ENABLE_TRM: False
          MULTICAST_GROUP_SUBNET: 239.1.1.0/25
          PHANTOM_RP_LB_ID1: 2
          PHANTOM_RP_LB_ID2: 3
          PHANTOM_RP_LB_ID3: 114
          PHANTOM_RP_LB_ID4: 115
```

### Debug Output

https://gist.github.com/skaszlik/0dfa70be57b145abbc941b4e18b8d3df
### Expected Behavior

The validation should allow the configuration

### Actual Behavior

[...]

PR #442, opened 2025-06-24
dcnm_fabric: Fix for #441

skaszlik:
The issue was in the parameter validation logic in **verify_playbook_params.py**

When comparing parameter values during validation, the code was inconsistently processing values through the conversion utilities:

- Playbook values (like RP_COUNT: 4) were processed through **make_boolean()**
- Controller values were processed through **make_boolean()**
- Default values from templates were not processed through conversions at all
- Rule values from templates were not processed through conversions
- This caused type mismatches during comparison where integer 4 from the playbook wouldn't properly compare with expected integer values in validation rules.

I modified the following methods in **verify_playbook_params.py** to ensure all values are consistently processed

### DIFF
--- a/plugins/module_utils/fabric/verify_playbook_params.py
+++ b/plugins/module_utils/fabric/verify_playbook_params.py
@@ -238,6 +238,13 @@ class VerifyPlaybookParams:
             msg += f"'value' not found in parameter {parameter} rule: {rule}"
             raise KeyError(msg)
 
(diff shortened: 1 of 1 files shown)


mikewiebe:
@skaszlik Is this ready to be reviewed?  I see it's still in draft status

allenrobel (approved):
Look's good!

mikewiebe (approved):

PR #444, opened 2025-06-25
dcnm_inventory: Device not reachable fixes

mikewiebe:
**Summary:**

The dcnm_inventory module does not currently handle cases where a device is unreachable.

This could be for any of the following reasons:
  * Device has been pre-provisioned
  * Device is in POAP mode
  * Device is not reachable due to some other network issue

**Adding devices to the fabric:**

When a device needs to be added to the fabric but is unreachable the following error message is generated and execution stops which is appropriate.

```python
msg = "Switch with IP {0} is not reachable or is not a valid IP".format(want_c["seedIP"])
```

If a device has already been added to the fabric but is currently unreachable it should still be managed because it's already part of the fabric.

This update makes an additional check to see if the device has already been added but is not reachable and does not error in this case.  The error message will still be generated for discovery when the device is not already part of the fabric.

**Removing devices from the fabric:**

Devices that have been pre-provisioned are not being removed properly so this also fixes that case.

### DIFF
--- a/plugins/modules/dcnm_inventory.py
+++ b/plugins/modules/dcnm_inventory.py
@@ -845,16 +845,31 @@ class DcnmInventory:
 
         self.diff_create = diff_create
 
+
     def get_diff_replace_delete(self):
 
         diff_delete = []
 
+        def check_have_c_in_want_list(have_c):
+            for want_c in self.want_create:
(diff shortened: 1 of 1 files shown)


ccoueffe (approved):
LGTM

Issue #424, opened 2025-06-03
dcnm_network replaced task to removes attachment succeeds but changed flag is set to false


neiljohn-gh:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Ansible Version and collection version

ansible [core 2.18.5]
collection version 3.8.0

### DCNM version

* V 12.2.2

### Affected module(s)

* dcnm_network

### Ansible Playbook

```yaml
- name: MERGED - Create, Attach Multiple Network with Single Switch Attach
  cisco.dcnm.dcnm_network:
    fabric: "fabric-name"
    state: merged
    config: "{{ conf_with_attachments }}"
  register: result
  tags: replaced

- name: REPLACED - Delete Attachments
  cisco.dcnm.dcnm_network: &conf1
    fabric: "fabric-name"
    state: replaced
    config: "{{ conf_without_attachments }}"
  register: result
  tags: replaced

# below fails
- name: ASSERT - Check if changed flag is true
  assert:
    that:
    - result.changed == true
  tags: replaced

```

### Debug Output

```json
"changed": false,
    "diff": [
        {
            "attach": [
                {
                    "deploy": false,
                    "ip_address": "192.168.x.x",
                    "ports": "Ethernet1/17,Ethernet1/18"
                }
            ],
            "net_name": "net1"
        }    
    ]
```

### Expected Behavior

changed should be true as attachments are removed and diff has correct contents

### Actual Behavior

changed is false

### Steps to Reproduce

1. Run a merge task for a network with attachments
2. Run a replaced task for the same network removing attachments

PR #448, opened 2025-06-30
Replaced changed flag #424

neiljohn-gh:
Fix for issue https://github.com/CiscoDevNet/ansible-dcnm/issues/424 (replaced state not showing correct changed flag when attachments are deleted)

Changelog:

- removed networks from diff_deploy if network deploy attribute is set to false in get_diff_replace
- added function to detach attachments and deploy the network before a delete to handle case when network is pending
- added retry count to wait_for_del_ready to prevent potential infinite loop (timeout is after 100 seconds)
- changed checking for del_ready from get network attachment to get network status
- fixed integration test templates to set deploy flag properly
- removed obsolete test case
- modified unit tests to match new endpoint


mikewiebe:
@neiljohn-gh Looks like there are conflicts that need to be resolved.  I will review once you address those.  Thanks!

Issue #455, opened 2025-07-02
dcnm_vrf: When a network is attached to a vrf, deletion fails without proper error handling


neiljohn-gh:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Affected module(s)

dcnm_vrf

* dcnm_XXXXX

### Ansible Playbook

```yaml
- name: MAIN - OVERRIDDEN - Create vrfs required for this test and remove all other vrfs
  cisco.dcnm.dcnm_vrf:
    fabric: "fabric"
    state: deleted
  tags: always 
```

### Debug Output

```
fatal: [nac-ndfc1]: FAILED! => {
    "changed": false,
    "module_stderr": "'NoneType' object is not subscriptable",
    "module_stdout": "",
    "msg": "MODULE FAILURE: No start of json char found\nSee stdout/stderr for the exact error"
} 
```
### Expected Behavior

Error should inform that networks are attached

### Actual Behavior

Fails without any information

### Steps to Reproduce

1. create a vrf 
2. attach a network
3. run deleted or override in the playbook

allenrobel:
This was resolved with #456, right?  If so, can we link it and close it?

By "link" I mean:

- i.e. either
  - Add `closes 456` to the initial comment in the PR, or
  - Use the right-pane "Development - Successfully merging this pull request may close these issues." to add this issue)
 
Thanks!

neiljohn-gh:
Fixed by #456

PR #456, opened 2025-07-02
dcnm_vrf now raises an error if networks are attached during deletion

neiljohn-gh:
Fix for #455 

When networks are attached to a vrf, deletion fails without proper information being raised.
This was because the check for networks was wrong. It was using the VRF switch attachments and not network attachments endpoint.

I moved the check for networks before any change is done to the vrf in case a vrf is being deleted and now it fails gracefully.

Changelog:

- added network attachment checks using a new endpoint
- added graceful error handling
- added a timeout in wait_for_vrf_del_ready of 100 seconds per vrf to prevent potential infinite loops
- removed the failure condition inside the wait_for_vrf_del_ready function as the failure is now being checked earlier
- updated Unit Tests

### DIFF
--- a/plugins/modules/dcnm_vrf.py
+++ b/plugins/modules/dcnm_vrf.py
@@ -583,6 +583,7 @@ dcnm_vrf_paths = {
         "GET_VRF_SWITCH": "/rest/top-down/fabrics/{}/vrfs/switches?vrf-names={}&serial-numbers={}",
         "GET_VRF_ID": "/rest/managed-pool/fabrics/{}/partitions/ids",
         "GET_VLAN": "/rest/resource-manager/vlan/{}?vlanUsageType=TOP_DOWN_VRF_VLAN",
+        "GET_NET_VRF": "/rest/resource-manager/fabrics/{}/networks?vrf-name={}"
     },
     12: {
         "GET_VRF": "/appcenter/cisco/ndfc/api/v1/lan-fabric/rest/top-down/fabrics/{}/vrfs",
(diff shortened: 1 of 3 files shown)


plugins/modules/dcnm_vrf.py
```diff
@@ -3758,6 +3760,18 @@ def push_to_remote(self, is_rollback=False):
...
         # created. This is needed specially for state: overridden
 
+        for vrf_name in self.diff_delete:
+            path = self.paths["GET_NET_VRF"].format(self.fabric, vrf_name)
+            resp = dcnm_send(self.module, "GET", path)
+            if resp.get("DATA") is None:
+                msg = f"Invalid Response from Controller. {resp}"
+                self.module.fail_json(msg=msg)
+            elif resp["DATA"] != []:
+                msg = "VRF has associated network attachments. "
```
mikewiebe, line 3770: Couple of minor suggestions:

Can you include the VRF name and fabric in the error message about attached networks?
* Example: f"VRF '{vrf_name}' in fabric '{self.fabric}' has associated network attachments."

Consider adding debug logging to show how many networks are attached to help with troubleshooting

plugins/modules/dcnm_vrf.py
```diff
@@ -3811,31 +3825,18 @@ def wait_for_vrf_del_ready(self, vrf_name="not_supplied"):
...
-                        msg += f"switch_name: {switch_name}, "
-                        msg += f"vlan_id: {vlan_id}"
-                        self.module.fail_json(msg=msg)
                     if attach["lanAttachState"] != "NA":
                         time.sleep(self.WAIT_TIME_FOR_DELETE_LOOP)
                         self.diff_delete.update({vrf: "DEPLOYED"})
                         ok_to_delete = False
                         break
                     self.diff_delete.update({vrf: "NA"})
+                if retry_count <= 0:
```
mikewiebe, line 3834: Wait Time Consistency:

The timeout uses self.WAIT_TIME_FOR_DELETE_LOOP for sleep but doesn't consider this in the retry count calculation

Consider making retry count dynamic based on wait time to ensure consistent timeout behavior

plugins/modules/dcnm_vrf.py
neiljohn-gh: addressed

plugins/modules/dcnm_vrf.py
neiljohn-gh: addressed

Issue #324, opened 2024-09-04
VRF module does not catch 'Failed' messages in the DATA field of the response from NDFC.


gnakh:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Ansible Version and collection version

ansible [core 2.17.2]
dcnm collection ver 3.4.3

### DCNM version

* 12.2.2.238

### Affected module(s)

* dcnm_vrf

### Ansible Playbook

```yaml
    - name: Create and attach VRFs to switches in {{ defaults.msd.fabric_name }} fabric
      cisco.dcnm.dcnm_vrf:
        fabric: '{{ defaults.msd.fabric_name }}'
        state: '{{ state | default("merged") }}'
        config: '{{ extra_vrfs }}'
      tags: vrfs
```

This is a sample VRF. Border leaves are 154.16 and 154.17
```yaml
extra_vrfs:
extra_vrfs:
  - vrf_name: iot
    vrf_id: 51200
    max_bgp_paths: 4
    vrf_template: '{{ defaults.vrf_template }}'
    vrf_extension_template: '{{ defaults.extension_vrf_template }}'
    vlan_id: 3200
    vrf_int_mtu: 9000
    #source: 123
    #service_vrf_template: null
    attach:
    - ip_address: 198.18.167.9
      deploy: true
    - ip_address: 198.18.167.8
      deploy: true
    - ip_address: 198.18.165.133 
      deploy: true           
    - ip_address: 198.18.167.25 
      deploy: true
    - ip_address: 198.18.167.19 
      deploy: true
    - ip_address: 198.18.154.16 
      vrf_lite:
        - peer_vrf: iot
          ipv4_addr: 172.16.12.2/30
          neighbor_ipv4: 172.16.12.1
          interface: Ethernet1/24
          dot1q: 200
          #extension_subint_mtu: 9000
        - peer_vrf: iot
          ipv4_addr: 172.16.12.10/30
          neighbor_ipv4: 172.16.12.9
          interface: Ethernet1/25
          dot1q: 200
          #extension_subint_mtu: 9000      
        - peer_vrf: iot
[...]

PR #457, opened 2025-07-02
dcnm_vrf: Handling "Fail" messages in Response Data from NDFC : Fix Issue #324

AKDRG:
Issue:
https://github.com/CiscoDevNet/ansible-dcnm/issues/324

Fix:
Inspecting the Response Data from the NDFC and checking if "fail" is part of the message. Marking "changed" and "fail" for the Task as "False" and "True", respectively.


allenrobel:
Hi Akshay, let's not make any changes to dcnm_vrf.py.  This file no longer exists after dcnm-vrf-pydantic-integration branch is merged.

If you'd like to checkout dcnm-vrf-pydantic-integration branch to apply your fixes there, let me know and I can stop work on it for you to make changes, and then pull your changes and continue work on it.

Thanks!
Allen

plugins/module_utils/network/dcnm/dcnm.py
```diff
@@ -859,3 +859,64 @@ def find_dict_in_list_by_key_value(search: list, key: str, value: str):
...
+    print(result)
+    # -> True
+
+    search_string = "not_found"
+    result = search_nested_json(content, search_string)
+    print(result)
+    # -> False
+    ```
+
+    """
```
mikewiebe, line 907: Can you double check.  Looks like this might have an early return bug with the early return on line `911` and line `918`.  If the search string is in any structure after the first one, it don't think it will be found.

The following slightly refactored version should fix that case but please test.

```python
def search_nested_json(obj, search_string):
    """Recursively search for string in nested structures"""
    if isinstance(obj, dict):
        for k, v in obj.items():
            if isinstance(v, (dict, list)):
                if search_nested_json(v, search_string):
                    return True
            elif isinstance(v, str) and search_string.lower() in v.lower():
                return True
    elif isinstance(obj, list):
        for item in obj:
            if search_nested_json(item, search_string):
                return True
    return False
```

AKDRG: Hi Mike,
Thanks for catching this! The additions of these return statements were a last minute thought and I was biased towards a test case, leading to this. Will take care to prevent these in future. Have rectified my mistake now and tested with various input data.

Thanks,
Akshay

mikewiebe (approved):

Issue #422, opened 2025-06-02
DCNM vPC Pair Module does not work on External fabric


peter8498:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Ansible Version and collection version

2.16.3

### DCNM version

12.2.3

### Affected module(s)

cisco.dcnm.dcnm_vpc_pair

### Ansible Playbook

```yaml
      - name: Merge VPC switch pair paremeters
        cisco.dcnm.dcnm_vpc_pair:
          src_fabric: "TEST-SVS-SL1"
          deploy: false
          state: merged
          config:
            - peerOneId: FDO2729038J
              peerTwoId: FDO27290391
              templateName: "vpc_pair"
              profile:
                ADMIN_STATE: True
                ALLOWED_VLANS: "all"
                DOMAIN_ID: 11
                FABRIC_NAME: TEST-SVS-SL1
                KEEP_ALIVE_HOLD_TIMEOUT: 3
                KEEP_ALIVE_VRF: management
                PC_MODE: active
                PEER1_DOMAIN_CONF: ""
                PEER1_KEEP_ALIVE_LOCAL_IP: 10.1.150.100
                PEER1_MEMBER_INTERFACES: e1/15,e1/16
                PEER1_PCID: 500
                PEER1_PO_CONF: "buffer-boost"
                PEER1_PO_DESC: "This is peer1 PC"
                PEER2_DOMAIN_CONF: ""
                PEER2_KEEP_ALIVE_LOCAL_IP: 10.1.150.101
                PEER2_MEMBER_INTERFACES: e1/15,e1/16
                PEER2_PCID: 500
                PEER2_PO_CONF: "buffer-boost"
                PEER2_PO_DESC: "This is peer2 PC"
```

### Debug Output

The full traceback is:
Traceback (most recent call last):
  File "/home/auto-user/.ansible/tmp/ansible-local-19789173qo7w29c/ansible-tmp-1748863609.5408928-1978969-171551319852259/AnsiballZ_dcnm_vpc_pair.py", line 107, in <module>
    _ansiballz_main()
[...]

PR #460, opened 2025-07-04
Resolving dcnm vpc pair keyError with ipAddress

sivakasi-cisco:


sivakasi-cisco:
#422

tests/integration/targets/dcnm_vpc_pair/tests/dcnm/dcnm_vpc_pair_merge.yaml
```diff
@@ -84,12 +84,12 @@
 
     - assert:
         that:
-          - 'result.changed == false'
+          - 'result.changed == true'
```
mikewiebe, line 87: Since this is an idempotence test, `result.changed` should be `false` not true.  The task is first run on line 30 above which creates state and the changed flag should be `true` but the second time it's already in the correct state so there should be no chane.

tests/integration/targets/dcnm_vpc_pair/tests/dcnm/dcnm_vpc_pair_override.yaml
```diff
@@ -138,12 +128,12 @@
 
     - assert:
         that:
-          - 'result.changed == false'
+          - 'result.changed == true'
```
mikewiebe, line 131: This is also an idempotence test so the flag should be false

tests/integration/targets/dcnm_vpc_pair/tests/dcnm/dcnm_vpc_pair_merge.yaml
```diff
@@ -84,12 +84,12 @@
...
           - '(result["diff"][0]["merged"] | length) == 0'
-          - '(result["diff"][0]["modified"] | length) == 0'
+          - '(result["diff"][0]["modified"] | length) == 1'
```
mikewiebe, line 89: Do we know why these lengths changed?

tests/integration/targets/dcnm_vpc_pair/tests/dcnm/dcnm_vpc_pair_merge.yaml
```diff
@@ -286,6 +285,8 @@
     - name: Create VPC switch pair without state and deploy flag
```
mikewiebe, line 285: You added the `deploy` and `state` flag here but it looks like the purpose of this test was to intentionally leave the state and deploy flag out and pick up the default values for those properties?

tests/integration/targets/dcnm_vpc_pair/tests/dcnm/dcnm_vpc_pair_replace.yaml
```diff
@@ -146,12 +136,12 @@
 
     - assert:
         that:
-          - 'result.changed == false'
+          - 'result.changed == true'
```
mikewiebe, line 139: Should be false.. idempotence test

tests/integration/targets/dcnm_vpc_pair/tests/dcnm/dcnm_vpc_pair_override.yaml
sivakasi-cisco: Addressed as above

tests/integration/targets/dcnm_vpc_pair/tests/dcnm/dcnm_vpc_pair_replace.yaml
sivakasi-cisco: Same as above

sivakasi-cisco:
Overall in the second changeset, I had removed all the changes made in the integration test yaml files. This is because, the actual problem lied with useVirtualPeerLink variable, as being set as null (or) None (or) False in different files. Hence, the variable set and the response obtained from NDFC are different, which shows a difference in comparison, leading to generating a diff in idemopotent tests.

Now, in the fix, the changes are made so that, the variable "useVirtualPeerLink" is set to false uniformly(being a boolean variable). Also the want and have comparison checks are updated.

plugins/module_utils/network/dcnm/dcnm_vpc_pair_utils.py
```diff
@@ -393,14 +393,21 @@ def dcnm_vpc_pair_compare_vpc_pair_objects(self, wobj, hobj):
...
-
-            # We found an object that matched all other key values, but differs in one of the params.
-            mismatch_reasons.append(
-                {key.upper() + "_MISMATCH": [wobj[key], hobj.get(key, None)]}
-            )
+        # Special handling for useVirtualPeerlink to treat None and False as equivalent
+        if key == "useVirtualPeerlink":
+            hval = hobj.get(key, None)
+            wval = wobj[key]
+            if not ((hval is None and wval is False) or (hval is False and wval is None)):
```
mikewiebe, line 400: Not sure I fully understand this if statement.  I wonder if it would simplify things to ensure `hval` and `wval` are booleans so you don't have to deal with None?

```python
hval = bool(hobj.get(key, None))
wval = bool(wobj[key])
```

sivakasi-cisco: Sure, Mike. Shall check this

Issue #461, opened 2025-07-04
dcnm_network module does not show an error when an unsupported parameter is in the playbook


neiljohn-gh:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Affected module(s)

dcnm_network

* dcnm_XXXXX

### Ansible Playbook

```yaml
    - name: Create Network
      cisco.dcnm.dcnm_network:
        fabric: neil-test
        state: merged
        config:
        - net_name: MYNETWORK1
          vrf_name: Ansible-prod
          is_l2only: False
          net_id: 30999
          vlan_id: 999
          gw_ip_subnet: 10.0.99.1/24
          vlan_name: MYNETWORK1
          int_desc: 
          mtu_l3intf: 9216
          dhcp_srvr1_ip: 1.1.1.1
          dhcp_srvr1_vrf: Ansible-prod
          dhcp_srvr2_ip: 2.2.2.2
          dhcp_srvr2_vrf: Ansible-prod
          dhcp_srvr3_ip: 3.3.3.3
          dhcp_srvr3_vrf: Ansible-prod
          dhcp_loopback_id: 2
          fake_parameter: fake_value
```

### Expected Behavior

throw an error stating fake_parameter is invalid

### Actual Behavior

Tasks executes while ignoring fake_parameter

### Steps to Reproduce

run playbook after creating the vrf

### References

related to #446

PR #464, opened 2025-07-07
Add extra parameters check when comparing a config to a spec

neiljohn-gh:
fix for #461 

added find_extra_params which returns a list of params in the config but not defined in the spec
eg:
```
    - name: Create Network
      cisco.dcnm.dcnm_network:
        fabric: fabric
        state: merged
        config:
        - net_name: MYNETWORK1
          vrf_name: Ansible-prod
          is_l2only: False
          net_id: 30999
          vlan_id: 999
          gw_ip_subnet: 10.0.99.1/24
          vlan_name: MYNETWORK1
          int_desc: 
          mtu_l3intf: 9216
          dhcp_srvr1_ip: 1.1.1.1
          dhcp_srvr1_vrf: Ansible-prod
          dhcp_srvr2_ip: 2.2.2.2
          dhcp_srvr2_vrf: Ansible-prod
          dhcp_srvr3_ip: 3.3.3.3
          dhcp_srvr3_vrf: Ansible-prod
          dhcp_loopback_id: 2
          fake_parameter: fake
```

fake_parameter will now throw an error

### DIFF
--- a/plugins/modules/dcnm_network.py
+++ b/plugins/modules/dcnm_network.py
@@ -550,6 +550,8 @@ class DcnmNetwork:
         else:
             self.paths = self.dcnm_network_paths[self.dcnm_version]
 
+        self.check_extra_params = True
+
         self.result = dict(changed=False, diff=[], response=[], warnings=[])
 
         self.failed_to_rollback = False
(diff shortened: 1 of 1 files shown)


plugins/modules/dcnm_network.py
```diff
@@ -2620,10 +2620,10 @@ def validate_input(self):
             if self.config:
                 msg = None
                 # Validate net params
-                valid_net, invalid_params = validate_list_of_dicts(self.config, net_spec)
+                valid_net, invalid_params = validate_list_of_dicts(self.config, net_spec, check_extra_params=True)
```
mikewiebe, line 2623: Can you make `check_extra_params` an instance variable instead that gets set and initialized to `True` in the `__init__` method.  That way we can set it in one place.  The just pass it in as `self.check_extra_params`

neiljohn-gh: updated

Issue #210, opened 2023-03-10
dcnm_vrf vrf_lite auto dot1q allocation doesn't work as expected


dsx1123:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Ansible Version and collection version

```
ansible [core 2.12.10]
  config file = /root/.ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /root/virtualenv/ansible/lib/python3.9/site-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /root/virtualenv/ansible/bin/ansible
  python version = 3.9.2 (default, Feb 28 2021, 17:03:44) [GCC 10.2.1 20210110]
  jinja version = 3.1.2
  libyaml = True
```

### DCNM version

* V 12.1.2e

### Affected module(s)

* dcnm_vrf

### Ansible Playbook

```yaml
# Copy-paste your anisble playbook here 
---
- name: VRF Lite issue
  hosts: ndfc2
  gather_facts: false
  tasks:
    - name: config vrf_lite
      cisco.dcnm.dcnm_vrf:
        fabric: fabric_n9kv
        config:
          - vrf_name: "{{item}}"
            attach:
              - ip_address: 192.168.123.13
                vrf_lite:
                - peer_vrf: "{{item}}"
                  interface: Ethernet1/1
                  ipv4_addr: 10.33.0.1/30
                  neighbor_ipv4: 10.33.0.2
      loop:
        - vrf_red
        - vrf_blue

```

### Debug Output

https://gist.github.com/dsx1123/afacb77fcd7ad257cd95239804549c7a
### Expected Behavior

second vrf vrf_blue should use 3 as dot1q tag
### Actual Behavior

playbook failed as second vrf is still using tag 2, NDFC reject the API

PR #467, opened 2025-07-10
Fix for DOT1Q_ID Auto Allocation Issue #210

AKDRG:
<<< Issue >>>
https://github.com/CiscoDevNet/ansible-dcnm/issues/210

<<< RCA >>>
DOT1Q_ID has been derived from the NDFC through the API:
appcenter/cisco/ndfc/api/v1/lan-fabric/rest/top-down/fabrics/<fabric-name>/vrfs/switches?vrf-names=<VRF NAME>&serial-numbers=<Switch Serial Number>

NDFC fills up the DOT1Q_ID as 2 in the response, even in case of multiple VRF attachments. 
```python
"extensionPrototypeValues":
{
"interfaceName": "Ethernet1/3",
"extensionType": "VRF_LITE",
"extensionValues": "{\"PEER_VRF_NAME\": \"\", \"NEIGHBOR_IP\": \"10.33.0.2\", \"VRF_LITE_JYTHON_TEMPLATE\": \"Ext_VRF_Lite_Jython\", \"enableBorderExtension\": \"VRF_LITE\", \"AUTO_VRF_LITE_FLAG\": \"false\", \"IP_MASK\": \"10.33.0.1/30\", \"MTU\": \"9216\", \"NEIGHBOR_ASN\": \"5\", \"IF_NAME\": \"Ethernet1/3\", \"IPV6_NEIGHBOR\": \"\", \"IPV6_MASK\": \"\", \"DOT1Q_ID\": \"2\", \"asn\": \"4001\"}",
"destInterfaceName": "Ethernet1/3",
"destSwitchName": "Spine4",
"destSerialNumber": "9ABCDFCCKEM7J"
}
```

<<< Fix Proposal >>>
Reserving DOT1Q ID for the VRF Lite Object through NDFC Resource Manager.
/appcenter/cisco/ndfc/api/v1/lan-fabric/rest/resource-manager/reserve-id

### DIFF
--- a/plugins/modules/dcnm_vrf.py
+++ b/plugins/modules/dcnm_vrf.py
@@ -3069,6 +3069,54 @@ class DcnmVrf:
 
         return extension_values_list
 
+    def get_vrf_lite_dot1q_id(self, serial_number: str, vrf_name: str, interface: str) -> int:
+        """
+        # Summary
+
+        Given a switch serial, vrf name and ifname, return the dot1q ID
+        reserved for the vrf_lite extension on that switch.
+
+        ## Raises
+
+        Calls fail_json if DNCM fails to reserve the dot1q ID.
+        """
+        caller = inspect.stack()[1][3]
+
+        msg = "ENTERED. "
+        msg += f"caller: {caller}. "
+        msg += f"serial_number: {serial_number}"
+        msg += f"vrf name: {vrf_name}"
+        msg += f"interface: {interface}"
+        self.log.debug(msg)
+
+        dot1q_id = None
+        path = "/appcenter/cisco/ndfc/api/v1/lan-fabric"
+        path += "/rest/resource-manager/reserve-id"
+        verb = "POST"
+        payload = {"scopeType": "DeviceInterface",
+                   "usageType": "TOP_DOWN_L3_DOT1Q",
+                   "serialNumber": serial_number,
+                   "ifName": interface,
+                   "allocatedTo": vrf_name}
+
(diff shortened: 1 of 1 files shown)


mikewiebe:
@AKDRG Here is the reference we spoke about for VRF-Lite

* https://github.com/netascode/ansible-dc-vxlan/issues/10
* Ensure manual allocation works for dot1q IDs
* Ensure auto allocation works for dot1q IDs

AKDRG:
<< Issues covered as part of this PR >>
1. "'NoneType' object is not subscriptable" Error while running the Playbook.
2.  While trying to use multiple interfaces in the VRF Lite object,  "string indices must be integers, not 'str'" error is thrown.
3.  Diff is getting populated while Idempotence Tests are run with VRF-Lite Objects.

<< RCA >>
1. VRFSegmentID is Auto-Populated in the NDFC, when the playbook is not supplying an VRFID. When an idempotence test runs with this type of config, VRF payloads with segmentID "null" is being pushed to the controller, causing the failure.
2. While processing the VRF Lite objects, the extension list was converted into a string prematurely, before the other items are added to the list, causing failure.
3. DOT1Q ID (currently being auto-populated from the controller), is differing in the have and want. 

<< Fix >>
1. "VRFSegmentID" key difference in the have and want, will be ignored if the segmentID is Null (VRF ID is not supplied by the user).
2. Fixed the conversion of the extension list to string, before all items get added. 
3. DOT1Q ID key difference in have and want, will be ignored if DOT1QID is None (not supplied by the user).

Thanks,
Akshay

mikewiebe (approved):

Issue #371, opened 2025-02-06
Add native vlan as an input to the interface module for any type of trunk port (trunk, po trunk etc)


mthurstocisco:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Currently when we are setting native vlan we are having to do it in the freeform section, would be nicer if we could use the defined box for this because then it flows down to the member interfaces in a nicer fashion (currently the freeform is on the member interfaces but it isn't visible anywhere in the GUI so can quite confusing)

### New or Affected modules(s):

* dcnm_interface

### DCNM version

* V 12.2.2

### Potential ansible task config

```yaml
# Copy-paste your ansible playbook
```

### References

**Additional context**

mthurstocisco:
@mikewiebe

neiljohn-gh:
#392 adds native vlan support.

PR #471, opened 2025-07-15
added ENABLE_LACP_VPC_CONV and DISABLE_LACP_SUSPEND to dcnm_interface

neiljohn-gh:
fix for #371 
Native lan was already added in another PR but the below fields are new
Added ENABLE_LACP_VPC_CONV and DISABLE_LACP_SUSPEND to dcnm_interface, under vpc profile.
Default value for both is false

### DIFF
--- a/plugins/modules/dcnm_interface.py
+++ b/plugins/modules/dcnm_interface.py
@@ -309,6 +309,17 @@ options:
             - Administrative state of the interface
             type: bool
             default: true
+          disable_lacp_suspend_individual:
+            description:
+            - If disabled, lacp will put the port to individual state and not suspend the port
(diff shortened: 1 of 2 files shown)


docs/cisco.dcnm.dcnm_fabric_module.rst
```diff
@@ -51,6 +51,2965 @@ Parameters
             </tr>
                                 <tr>
                     <td class="elbow-placeholder"></td>
+                <td colspan="2">
```
mikewiebe, line 54: Looks like the documentation prep collection was not run after a bunch of new properties were added to the fabric module.  Good that they are being added here.

mikewiebe (approved):

PR #472, opened 2025-07-16
ConversionUtils() unit tests and hardening

allenrobel:
# Summary

The `ConversionUtils` class lacked associated unit tests.  Adding them now.

## Hardening

- `translate_mac_address` could throw an unintended TypeError with integer input since the re.sub() was trying to alter the passed `mac_addr`, which could potentially be an `int`.  Fixed by casting `mac_addr` to `str` within the `re.sub()`.

## Usability

- `translate_mac_address` was returning the altered MAC address in its ValueError which could be confusing to the end user.  Fixed by copying the original `mac_addr` and using the copy in the ValueError.

## Updates to other unit tests

Two fabric unit tests started failing after the above changes now that we are using the original MAC address in the ValueError.  Fixed the asserts in these to expect the original MAC address.

### DIFF
--- a/tests/unit/module_utils/common/test_conversion_utils.py
+++ b/tests/unit/module_utils/common/test_conversion_utils.py
@@ -281,6 +281,6 @@ def test_conversion_utils_00050(value, expected, raises) -> None:
         with does_not_raise():
             assert instance.translate_mac_address(value) == expected
     else:
-        match = f"Invalid MAC address: {value}"
-        with pytest.raises(ValueError, match=re.escape(match)):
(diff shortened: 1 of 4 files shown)


tests/unit/module_utils/common/test_conversion_utils.py
```diff
@@ -0,0 +1,352 @@
+# Copyright (c) 2025 Cisco and/or its affiliates.
```
mikewiebe, line 1: Wondering ... Do we want to pick up and run these tests in our GitHub actions?

PR #484, opened 2025-07-31
Fix issue with breakout with multiple switch

ccoueffe:


plugins/modules/dcnm_interface.py
```diff
@@ -3852,6 +3862,19 @@ def dcnm_intf_get_have_all_with_sno(self, sno):
         if resp and "DATA" in resp and resp["DATA"]:
             self.have_all.extend(resp["DATA"])
 
+    def dcnm_intf_get_have_all_breakout_interfaces(self, sno):
+        # This function will get policies for a given serial number and
+        # populate the breakout interfaces in self.have_breakout.
+        path = "/appcenter/cisco/ndfc/api/v1/lan-fabric/rest/control/policies/switches/{}".format(sno)
+        resp = dcnm_send(self.module, "GET", path)
```
mikewiebe, line 3869: @ccoueffe Can you get me the exact output that is saved to this `resp` variable?  I will need this to fix the unit tests.

mikewiebe (approved):

Issue #417, opened 2025-05-19
dcnm_network: 500 error - networkName missing in networkTemplateConfig


dalamanster:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Ansible Version and collection version
                                   
* ansible [core 2.18.5]

### DCNM version

* Version 3.2.2f
* Version 3.2.1

### Affected module(s)

* dcnm_network

### Ansible Playbook

```yaml
    - name: Create and attach network in Enhanced Classic fabric
      cisco.dcnm.dcnm_network:
        fabric: DC4PRP
        state: merged
        config:
        - net_name: test_nw_999
          vlan_id: 2222
          vlan_name: test_nw_999
          net_template: Network_Classic
          multicast_group_address: ""
          is_l2only: True
          attach:
          - ip_address: "10.17.23.32"
            ports: []
          deploy: False
```

### Expected Behavior

dcnm network created

### Actual Behavior

500 Internal Server Error

fatal: [ndfcdev]: FAILED! => {"changed": false, "msg": {"DATA": {"Error": "Internal Server Error", "message": "An unexpected error occurred during template execution. Please retry after some time. Please contact Cisco TAC if the issue persists.", "path": "/rest/top-down/fabrics/DC4PRP/networks", "status": "500", "timestamp": "2025-05-19 08:33:44.863"}, "MESSAGE": "Internal Server Error", "METHOD": "POST", "REQUEST_PATH": "https://10.17.22.30:443/appcenter/cisco/ndfc/api/v1/lan-fabric/rest/top-down/fabrics/DC4PRP/networks", "RETURN_CODE": 500, "ROLLBACK_RESULT": "SUCCESS - Attempted rollback of the task has succeeded"}}

### Possible issue

I made diff between:
- API post request made by ansible-dcnm script (ansible task run)
- API post requests made from NDFC apidocs webpage (manual run of api call to create the network)

Key "networkName" is missing in the "networkTemplateConfig" field.

When I add this in the code - the playbook works as expected. 

* POST REQUEST (before change):

[...]

PR #485, opened 2025-08-05
fix for #417 (dcnm_network.py - networkName missing in networkTemplateConfig)

dalamanster:
This is proposed fix for #417 .

Please somebody check it before merge. This is my first PR for this project. I put "netName" and "networkName" variables on all places where also other network parameters were defined/altered. 

This helped me to fix the #417 and I am not getting 500 error anymore.


mtarking:
Thanks @dalamanster for the contribution. 

@mikewiebe we do need to double check as I think there's a subtle delta here for MSD fabrics.

PR #504, opened 2025-09-04
fix key issue: 'ENABLE_MACSEC' on dcnm_links.py 

ccoueffe:
Add safe.get for ENABLE_MACSEC, as the key is not always returned by ND.

### DIFF
--- a/plugins/modules/dcnm_links.py
+++ b/plugins/modules/dcnm_links.py
@@ -3075,8 +3075,9 @@ class DcnmLinks:
                 == self.templates["int_intra_fabric_unnum_link"]
             )
         ):
-            if (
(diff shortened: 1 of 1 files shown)


plugins/modules/dcnm_links.py
```diff
@@ -3076,7 +3076,8 @@ def dcnm_links_compare_intra_fabric_link_params(self, wlink, hlink):
             )
         ):
             if (
-                str(wlink["nvPairs"]["ENABLE_MACSEC"]).lower()
+                hlink["nvPairs"].get("ENABLE_MACSEC") and
```
mikewiebe, line 3079: I understand updating to use a safe `get` but why the additional `and` check here?

Issue #505, opened 2025-09-04
DCNM_VRF module shouldn't alocate vlan if l3vni_wo_vlan is used


mkilar123:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Ansible Version and collection version

ansible [core 2.18.8]
  config file = /Users/mkilar/Desktop/Projects/Vxlan_as_Code/nac-ndfc/ansible.cfg
  configured module search path = ['/Users/mkilar/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/ansible
  ansible collection location = /Users/mkilar/.pyenv/versions/sac-ndfc/lib/python3.12/site-packages/ansible/collections:/Users/mkilar/Desktop/Projects/Vxlan_as_Code/nac-ndfc/collections/ansible_collections
  executable location = /Library/Frameworks/Python.framework/Versions/3.12/bin/ansible
  python version = 3.12.4 (v3.12.4:8e8a4baf65, Jun  6 2024, 17:33:18) [Clang 13.0.0 (clang-1300.0.29.30)] (/Library/Frameworks/Python.framework/Versions/3.12/bin/python3.12)
  jinja version = 3.1.5
  libyaml = True

### DCNM version

* NDFC 3.2.1

### Affected module(s)
* dcnm_vrf

### Ansible Playbook

```yaml
- hosts: DB_BGP_CALO
  any_errors_fatal: true
  gather_facts: no
  collections:
    - cisco.dcnm
  
  tasks:
    - name: Create vrf
      cisco.dcnm.dcnm_vrf:
        fabric: "BGP_CALO"
        state: merged
        config:
        - vrf_name: vrf_ansible_test
          vrf_id: 9008011
          vrf_template: Default_VRF_Universal
          vrf_extension_template: Default_VRF_Extension_Universal
          vrf_int_mtu: 9216
          vrf_intf_desc: Configured by Ansible NetAsCode
          vrf_description: Configured
          trm_enable: false
          static_default_route: true
[...]

PR #508, opened 2025-09-16
Fix for L3VNI W/O VLAN_ID generation #505

AKDRG:
<< Issue >>
#505 

<< RCA >>
When a VRF (with L3VNI W/O enabled) is created using DCNM_VRF, vlan_id which is user specified or autogenerated from the module is sent to the NDFC in the payload. However, NDFC doesn't ignore it and it is set for the VRF.  
Note: The vlan-id is not used when VRF configuration is pushed to the device, and is not allocated as a resource by the NDFC and can be still reused. Also, vlan autogeneration skips the ID and moves on to the next one, thinking it's allocated. This causes inconsistency. 

<< Fix >>
Preventing autogeneration and transmitting of user-specified VLAN for a VRF with L3VNI W/O VLAN setting enabled. 
Also, deriving the default setting value of L3VNI W/O VLAN at the VRF level from the Fabric level setting, instead of "False".


mikewiebe:
@AKDRG Tested the latest updates and it's looking much better.  There is still one scenario that is not working quite right.

* Create a VRF with the following playbook
    * `l3vni_wo_vlan` is `true`
    * vlan is commented out

```yaml
    - name: Manage VRF with 'l3vni_wo_vlan' Property
      cisco.dcnm.dcnm_vrf:
        fabric: nac-fabric1
        state: replaced
        config:
        - vrf_name: vrf_two
          vrf_id: 9008012
          l3vni_wo_vlan: true
          # vlan: 5000
          vrf_template: Default_VRF_Universal
          vrf_extension_template: Default_VRF_Extension_Universal
          vrf_int_mtu: 9216
          vrf_intf_desc: Configured by Ansible NetAsCode 64
          vrf_description: Configured
          trm_enable: false
          static_default_route: true
          no_rp: false
          netflow_enable: false
          ipv6_linklocal_enable: true
          disable_rt_auto: false
          deploy: false
          attach:
            - ip_address: 192.168.84.55
            - ip_address: 192.168.84.56

This will create a VRF without a vlan as expected

* Next modify the `l3vni_wo_vlan` to `false` and provide a vlan

```yaml
    - name: Manage VRF with 'l3vni_wo_vlan' Property
      cisco.dcnm.dcnm_vrf:
        fabric: nac-fabric1
        state: replaced
        config:
        - vrf_name: vrf_two
          vrf_id: 9008012
          l3vni_wo_vlan: false
          vlan: 5000
          vrf_template: Default_VRF_Universal
          vrf_extension_template: Default_VRF_Extension_Universal
          vrf_int_mtu: 9216
          vrf_intf_desc: Configured by Ansible NetAsCode 64
          vrf_description: Configured
          trm_enable: false
          static_default_route: true
          no_rp: false
          netflow_enable: false
          ipv6_linklocal_enable: true
          disable_rt_auto: false
          deploy: false
          attach:
            - ip_address: 192.168.84.55
            - ip_address: 192.168.84.56
```

Instead of the vlan getting set to `5000` in this case it appears to be getting a vlan from the ND resource pool.

Screenshot 2025-10-23 at 3 11 43 PM

PR #532, opened 2025-10-24
Multisite (MSD) Network support for DCNM_NETWORK Module

neiljohn-gh:
Added msd support for dcnm_network.
child_fabric_config key enables you to pass the child specific network attributes to the child fabrics.
Example:
```yaml
- name: Create network on MSD parent fabric with child fabric configuration
  cisco.dcnm.dcnm_network:
    fabric: msd-parent-fabric
    state: merged
    config:
    - net_name: ansible-test-network
      vrf_name: NaC-VRF01
      net_template: Default_Network_Universal
      net_extension_template: Default_Network_Extension_Universal
      net_id: 130001
      vlan_id: 2301
      vlan_name: NaC-Net01_vlan2301
      gw_ip_subnet: '192.168.12.1/24'
      routing_tag: 1234
      int_desc: "Configured by Ansible NetAsCode"
      mtu_l3intf: 9216
      arp_suppress: false
      route_target_both: false
      is_l2only: false
      # Child fabric specific configurations
      child_fabric_config:
        - fabric: child-fabric-1
          # Child-specific parameters not allowed on parent fabrics
          l3gw_on_border: true
          dhcp_loopback_id: 204
          multicast_group_address: '239.1.1.1'
        - fabric: child-fabric-2
          # Different configuration for second child
          l3gw_on_border: false
          dhcp_loopback_id: 205
          # deploy not specified - inherits from parent
      # Attachments are only configured on parent fabric, not in child_fabric_config
      attach:
        - ip_address: 192.168.10.203
          ports: [Ethernet1/13, Ethernet1/14]
        - ip_address: 192.168.10.204
          ports: [Ethernet1/13, Ethernet1/14]
      deploy: true  # Parent deploy setting, inherited by children unless overridden
```

Internally, an action plugin splits this into multiple configs, one for each fabric mentioned and then pushes these configs one by one into the traditional base model.

changelog:
- Added action plugin that executes base module based on fabric type (multisite_parent, multisite_child, standalone)
- Added scoping of attributes in case of fabric_type in the base module
- Added IT and UT

### DIFF
--- a/plugins/modules/dcnm_network.py
+++ b/plugins/modules/dcnm_network.py
@@ -21,10 +21,10 @@ __author__ = "Chris Van Heuveln, Shrishail Kariyappanavar, Karthik Babu Harichan
 DOCUMENTATION = """
 ---
 module: dcnm_network
-short_description: Add and remove Networks from a DCNM managed VXLAN fabric.
+short_description: Add and remove Networks from a ND managed VXLAN fabric.
 version_added: "0.9.0"
 description:
-    - "Add and remove Networks from a DCNM managed VXLAN fabric."
+    - "Add and remove Networks from a ND managed VXLAN fabric."
     - "For multisite (MSD) fabrics, child fabric configurations can be specified using the child_fabric_config parameter"
     - "The attribute _fabric_type (standalone, multisite_parent, multisite_child) is automatically detected and should not be manually specified by the user"
 author: Chris Van Heuveln(@chrisvanheuveln), Shrishail Kariyappanavar(@nkshrishail) Praveen Ramoorthy(@praveenramoorthy)
@@ -46,7 +46,7 @@ options:
     choices: ['multisite_child', 'standalone', 'multisite_parent']
   state:
     description:
-    - The state of DCNM after module completion.
+    - The state of ND after module completion.
     type: str
     choices:
       - merged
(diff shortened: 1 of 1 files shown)


mikewiebe:
Few merge conflicts need to be resolved after merging the recent update to this module to include new dhcp parameters

tests/unit/modules/dcnm/fixtures/dcnm_network.json
```diff
@@ -8,6 +8,46 @@
...
     },
+    "fabric_associations": {
+        "RETURN_CODE": 200,
+        "METHOD": "GET",
+        "REQUEST_PATH": "https://mock-ndfc/appcenter/cisco/ndfc/api/v1/lan-fabric/rest/control/fabrics/msd/fabric-associations",
+        "MESSAGE": "OK",
+        "DATA": [
+            {
+                "fabricId": 1,
+                "fabricName": "test_network",
```
mikewiebe: Just a minor thing but can we change the name of this fabric to be "standalone_fabric" instead of "test_network".  I know we have the other info below that makes it clear it's a standalone fabric but just visually easier to read when debugging issues in the future.

plugins/action/dcnm_network.py
```diff
@@ -0,0 +1,586 @@
...
+        if not state:
+            result['failed'] = True
+            result['msg'] = "The 'state' parameter is required"
+            return result
+
+        config = module_args.get('config')
+
+        if not isinstance(config, list) or not config:
+            # For 'query' and 'deleted', allow empty config (interpreted as all networks)
+            if state in ['query', 'deleted']:
```
mikewiebe: We need to support `[]` for state `overridden` also. When we pass an empty list it should remove all networks

PR #553, opened 2025-11-14
DcnmBootflash: type hints, class property RestSend

allenrobel:
This commit addresses technical debt in the main dcnm_bootflash.py module.

1. Adds type hints
2. Removes class decorator @Properties.add_rest_send in favor of a class property for RestSend
3. Run black linter which caused formatting changes due to relaxation of line-length constraints

Per Copilot's review comments, I'm modifying the claim that "no functional changes" exist in this PR.  In fact, the changes in this PR introduce subtle (but desired) behavioral changes.

These changes are related to:

1. Handling of "falsey" values (e.g. "", []) vs None.  Since we are type-hinting and instantiating variables to str, list, etc, rather than e.g. Union[str, None], our conditionals need to change accordingly.

3. A small "fix" related to handling the case where task is None in the error path in main().

### DIFF
--- a/plugins/modules/dcnm_bootflash.py
+++ b/plugins/modules/dcnm_bootflash.py
@@ -209,7 +209,7 @@ class Common:
 
     def __init__(self, params: dict[str, Any]) -> None:
         self.class_name = self.__class__.__name__
-        method_name = inspect.stack()[0][3]
+        method_name: str = inspect.stack()[0][3]
 
         self.log = logging.getLogger(f"dcnm.{self.class_name}")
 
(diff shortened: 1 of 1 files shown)


plugins/modules/dcnm_bootflash.py
```diff
@@ -361,29 +350,69 @@ def raise_type_error(msg):
...
+        """
+        if not self._rest_send.params:
+            msg = f"{self.class_name}.rest_send: "
+            msg += "rest_send has not been set (missing params)."
+            raise ValueError(msg)
+        return self._rest_send
+
+    @rest_send.setter
+    def rest_send(self, value: RestSend):
+        method_name = inspect.stack()[0][3]
```
mikewiebe: I see other places where you are using type hints for `method_name`.  Is there a reason you are not doing it here?

PR #562, opened 2025-11-20
Pydantic mocks: shared resource files for 3rd-party imports

allenrobel:
# Summary

This PR introduces a shared location for files which satisfy 3rd-party import requirements for Ansible sanity import tests.

It uses Pydantic as an initial proof-of-concept.

The location can be used for other 3rd-party import guard mocks, like requests, etc.

Once this is merged, we can proceed to leverage it to import mocks for Pydantic objects in all other files that currently use explicit local definitions for Pydantic objects within their import guards.

## Note to reviewers

I've unleashed Copilot as a reviewer so that you don't have to spend time of mundane changes.

### module_utils/common/third_party/pydantic.py

Initial set of mocks for Pydantic.

### modules/dcnm_fabric_group.py

This module doesn't actually use Pydantic (doh!).  Removed the import guards.  No other functional changes for this file.

### plugins/action/test/integration/ndfc_inventory_validate.py

This is the file we have tested by replacing its import guards with imported guards from module_utils/common/third_party/pydantic.py.

This is working, so I've removed the explicit definitions in this file.

### DIFF
--- a/plugins/action/tests/integration/ndfc_inventory_validate.py
+++ b/plugins/action/tests/integration/ndfc_inventory_validate.py
@@ -11,22 +11,6 @@ try:
 except ImportError:
     HAS_PYDANTIC = False
     from ....module_utils.common.third_party.pydantic import BaseModel, model_validator, validator, ValidationError
-    # # Create dummy classes to allow import without pydantic
-    # BaseModel = object
-
-    # def model_validator(**kwargs):
-    #     """Dummy model_validator decorator that accepts any arguments"""
-    #     def decorator(func):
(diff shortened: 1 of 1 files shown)


allenrobel:
> * [nitpick] The `traceback` module is imported but is now only used once (line 1206). Since Pydantic-related traceback usage was removed, verify if this import is still necessary or if it can be moved closer to its usage.
> 
> ```
> import traceback
> ```
> 

This is a valid point.  But this should be done in a separate PR.  I'll open a new PR to clean this up.

PR #589, opened 2025-12-11
Remove vim swap file from repository

sivakasi-cisco:


sivakasi-cisco:
Got this changes covered in 551 lint changes

PR #601, opened 2026-01-05
Add new 'dhcp_servers' key to allow up to 16 DHCP servers on overlay networks

miguecor:
This PR adds a new 'dhcp_servers' key under overlay networks that allows up to 16 DHCP servers to be configured.


miguecor:
@mikewiebe and @allenrobel: This PR is ready for review.

PR #612, opened 2026-01-13
Add xconnect dcnm network

### LINKED ISSUE #604: ND4.1: dcnm_network - Add xconnect parameter
### Community Note

### Description

Add new key in 4.1 xconnect.

Image

Example:

Method: PUT
/appcenter/cisco/ndfc/api/v1/lan-fabric/rest/top-down/fabrics/<fabric>/networks/<MyNetwork>

```json
{
    "networkTemplateConfig": {
        "gatewayIpAddress": "",
        "gatewayIpV6Address": "",
        "vlanName": "",
        "intfDescription": "Configured by Ansible NetAsCode",
        "mtu": "9216",
        "secondaryGWs": "",
        "dsVni": "",
        "type": "Normal",
        "dhcpServerAddr1": "",
        "vrfDhcp": "",
        "dhcpServerAddr2": "",
        "vrfDhcp2": "",
        "dhcpServerAddr3": "",
        "vrfDhcp3": "",
        "secondaryGW1": "",
        "secondaryGW2": "",
        "secondaryGW3": "",
        "secondaryGW4": "",
        "suppressArp": "false",
        "enableIR": "false",
        "xconnect": true,              <<<<< New key to manage xconnect parameter
        "mcastGroup": "239.0.0.2",
        "dhcpServers": "",
        "loopbackId": "",
        "tag": "12345",
        "trmEnabled": "false",
        "igmpVersion": "",
        "trmV6Enabled": "",
        "rtBothAuto": "false",
        "ENABLE_NETFLOW": "false",
        "SVI_NETFLOW_MONITOR": "",
        "VLAN_NETFLOW_MONITOR": "",
        "enableL3OnBorder": "false",
        "vlanId": "1002",
        "segmentId": 10002,
        "vrfName": "NA",
        "nveId": "1",
        "isLayer2Only": true,
        "str1": "",
        "str2": "",
        "flagSet": "",
        "isIpDhcpRelay": "",
        "isIp6DhcpRelay": "",
        "localAsn": "",
        "dsvniRouteMapName": "",
        "dsvniRouteMapSeqNum": "",
        "dsvniExtComListName": "",
        "dsvniExtComSeqNum": "",
        "createSvi": "",
        "switchRole": "",
        "isSharedBorder": "",
        "rtAsn": ""
    }
}
```

### New or Affected modules(s):

* dcnm_network

### DCNM version

* V 12.4

### Potential ansible task config

```yaml
# Copy-paste your ansible playbook
```

### References

**Additional context**

ccoueffe:
fix: #604 

Add xconnect for ND 4.1:

```
    - cisco.dcnm.dcnm_network:
        state: merged
        fabric: myFab
        config:
          - net_name: TEST-XCONNECT-NET1
            is_l2only: true
            vlan_id: 2001
            xconnect: true
```

```
changed: [myFab] => {
    "changed": true,
    "diff": [
        {
            "arp_suppress": false,
            "attach": [],
            "dhcp_loopback_id": "",
            "dhcp_servers": "",
            "dhcp_srvr1_ip": "",
            "dhcp_srvr1_vrf": "",
            "dhcp_srvr2_ip": "",
            "dhcp_srvr2_vrf": "",
            "dhcp_srvr3_ip": "",
            "dhcp_srvr3_vrf": "",
            "gw_ip_subnet": "",
            "gw_ipv6_subnet": "",
            "int_desc": "",
            "intfvlan_nf_monitor": "",
            "is_l2only": true,
            "l3gw_on_border": false,
            "mtu_l3intf": "",
            "multicast_group_address": "239.1.1.1",
            "net_extension_template": "Default_Network_Extension_Universal",
            "net_id": 30000,
            "net_name": "TEST-XCONNECT-NET1",
            "net_template": "Default_Network_Universal",
            "netflow_enable": false,
            "route_target_both": false,
            "secondary_ip_gw1": "",
            "secondary_ip_gw2": "",
            "secondary_ip_gw3": "",
            "secondary_ip_gw4": "",
            "trm_enable": false,
            "vlan_id": 2001,
            "vlan_name": "",
            "vlan_nf_monitor": "",
            "vrf_name": "NA",
            "xconnect": true
        }
    ],
    "invocation": {
        "module_args": {
            "config": [
                {
                    "is_l2only": true,
                    "net_name": "TEST-XCONNECT-NET1",
                    "vlan_id": 2001,
                    "xconnect": true
                }
            ],
            "fabric": "myFab",
            "state": "merged"
        }
    },
    "response": [
        {
            "DATA": {},
            "MESSAGE": "OK",
            "METHOD": "PUT",
            "REQUEST_PATH": "https://x.x.x.x:443/appcenter/cisco/ndfc/api/v1/lan-fabric/rest/top-down/fabrics/myFab/networks/TEST-XCONNECT-NET1",
            "RETURN_CODE": 200
        }
    ]
}

PLAY RECAP *********************************************************************************************************************************************************************************************************************
myFab                 : ok=1    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
```

### DIFF
--- a/plugins/modules/dcnm_network.py
+++ b/plugins/modules/dcnm_network.py
@@ -306,6 +306,12 @@ options:
         - Not applicable at Multisite parent fabric level
         type: str
         required: false
+      xconnect:
+        description:
+        - Enable XConnect feature. Only available for ND version > 4.1.
+        type: bool
+        default: false
+        required: false
       attach:
         description:
         - List of network attachment details
@@ -1001,6 +1007,11 @@ from ansible_collections.cisco.dcnm.plugins.module_utils.network.dcnm.dcnm impor
     validate_list_of_dicts,
     sanitize_lan_attach_list
 )
+from ..module_utils.common.controller_version_v2 import ControllerVersion
+from ..module_utils.common.rest_send_v2 import RestSend
+from ..module_utils.common.response_handler import ResponseHandler
+from ..module_utils.common.sender_dcnm import Sender
+from ..module_utils.common.exceptions import ControllerResponseError
 
 
 class DcnmNetwork:
@@ -1102,6 +1113,8 @@ class DcnmNetwork:
 
         msg = f"self.dcnm_version: {self.dcnm_version}"
         self.log.debug(msg)
+
+        self.ndfc_version = self._get_ndfc_version()
 
         self.inventory_data = get_nd_fabric_inventory_details(self.module, self.dcnm_version, self.fabric, self.fabric_details)
 
@@ -1622,6 +1635,7 @@ class DcnmNetwork:
         nf_en_changed = False
         intvlan_nfmon_changed = False
         vlan_nfmon_changed = False
+        xconnect_changed = False
 
         if want.get("networkId") and want["networkId"] != have["networkId"]:
             self.module.fail_json(msg="networkId can not be updated on existing network: {0}".format(want["networkName"]))
@@ -1679,6 +1693,8 @@ class DcnmNetwork:
         intvlan_nfen_have = json_to_dict_have.get("SVI_NETFLOW_MONITOR", "")
         vlan_nfen_want = json_to_dict_want.get("VLAN_NETFLOW_MONITOR", "")
         vlan_nfen_have = json_to_dict_have.get("VLAN_NETFLOW_MONITOR", "")
+        xconnect_want = str(json_to_dict_want.get("xconnect", "")).lower()
+        xconnect_have = str(json_to_dict_have.get("xconnect", "")).lower()
 
         if vlanId_have != "":
             vlanId_have = int(vlanId_have)
@@ -1796,6 +1812,10 @@ class DcnmNetwork:
                 net_name_diff = net_name_have != net_name_want
                 comparisons.append(net_name_diff)
 
+            if "xconnect" not in skipped_template_keys:
+                xconnect_diff = xconnect_have != xconnect_want
+                comparisons.append(xconnect_diff)
+
             if any(comparisons):
                 # The network updates with missing networkId will have to use existing
                 # networkId from the instance of the same network on DCNM.
@@ -1848,6 +1868,9 @@ class DcnmNetwork:
                         intvlan_nfmon_changed = True
                     if vlan_nfen_have != vlan_nfen_want:
                         vlan_nfmon_changed = True
+                if self._ndfc_version_gte("12.4.1"):
+                    if xconnect_have != xconnect_want:
+                        xconnect_changed = True
 
                 want.update({"networkId": have["networkId"]})
                 create = want
@@ -1957,6 +1980,10 @@ class DcnmNetwork:
                 net_name_diff = net_name_have != net_name_want
                 comparisons.append(net_name_diff)
 
+            if "xconnect" not in skipped_template_keys:
+                xconnect_diff = xconnect_have != xconnect_want
+                comparisons.append(xconnect_diff)
+
             if any(comparisons):
                 # The network updates with missing networkId will have to use existing
                 # networkId from the instance of the same network on DCNM.
@@ -2006,6 +2033,9 @@ class DcnmNetwork:
                         intvlan_nfmon_changed = True
                     if vlan_nfen_have != vlan_nfen_want:
                         vlan_nfmon_changed = True
+                if self._ndfc_version_gte("12.4.1"):
+                    if xconnect_have != xconnect_want:
+                        xconnect_changed = True
 
                 want.update({"networkId": have["networkId"]})
                 create = want
@@ -2035,6 +2065,7 @@ class DcnmNetwork:
             nf_en_changed,
             intvlan_nfmon_changed,
             vlan_nfmon_changed,
+            xconnect_changed,
         )
 
     def update_create_params(self, net):
@@ -2111,6 +2142,8 @@ class DcnmNetwork:
             template_conf.update(ENABLE_NETFLOW=net.get("netflow_enable", False))
             template_conf.update(SVI_NETFLOW_MONITOR=net.get("intfvlan_nf_monitor", ""))
             template_conf.update(VLAN_NETFLOW_MONITOR=net.get("vlan_nf_monitor", ""))
+        if self._ndfc_version_gte("12.4.1"):
+            template_conf.update(xconnect=net.get("xconnect", False))
 
         if template_conf["vlanId"] is None:
             template_conf["vlanId"] = ""
@@ -2264,6 +2297,8 @@ class DcnmNetwork:
                     t_conf.update(ENABLE_NETFLOW=json_to_dict.get("ENABLE_NETFLOW", False))
                     t_conf.update(SVI_NETFLOW_MONITOR=json_to_dict.get("SVI_NETFLOW_MONITOR", ""))
                     t_conf.update(VLAN_NETFLOW_MONITOR=json_to_dict.get("VLAN_NETFLOW_MONITOR", ""))
+                if self._ndfc_version_gte("12.4.1"):
+                    t_conf.update(xconnect=json_to_dict.get("xconnect", False))
 
                 # Remove mcastGroup when Fabric is MSD
                 if "mcastGroup" not in json_to_dict:
@@ -2319,6 +2354,8 @@ class DcnmNetwork:
                             t_conf.update(ENABLE_NETFLOW=json_to_dict.get("ENABLE_NETFLOW", False))
                             t_conf.update(SVI_NETFLOW_MONITOR=json_to_dict.get("SVI_NETFLOW_MONITOR", ""))
                             t_conf.update(VLAN_NETFLOW_MONITOR=json_to_dict.get("VLAN_NETFLOW_MONITOR", ""))
+                        if self._ndfc_version_gte("12.4.1"):
+                            t_conf.update(xconnect=json_to_dict.get("xconnect", False))
 
                         l2net.update({"networkTemplateConfig": json.dumps(t_conf)})
                         del l2net["displayName"]
@@ -2845,6 +2882,7 @@ class DcnmNetwork:
         nf_en_changed = {}
         intvlan_nfmon_changed = {}
         vlan_nfmon_changed = {}
+        xconnect_changed = {}
 
         for want_c in self.want_create:
             found = False
@@ -2877,6 +2915,7 @@ class DcnmNetwork:
                         nf_en_chg,
                         intvlan_nfmon_chg,
                         vlan_nfmon_chg,
+                        xconnect_chg,
                     ) = self.diff_for_create(want_c, have_c)
 
                     gw_changed.update({want_c["networkName"]: gw_chg})
@@ -2902,6 +2941,7 @@ class DcnmNetwork:
                     nf_en_changed.update({want_c["networkName"]: nf_en_chg})
                     intvlan_nfmon_changed.update({want_c["networkName"]: intvlan_nfmon_chg})
                     vlan_nfmon_changed.update({want_c["networkName"]: vlan_nfmon_chg})
+                    xconnect_changed.update({want_c["networkName"]: xconnect_chg})
                     if diff:
                         diff_create_update.append(diff)
                     break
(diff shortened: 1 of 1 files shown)


allenrobel (commented):
This looks fine to me.  The only minor points would be:

1. We might want to add integration test for xconnect.
2. Since there is no 4.x version check around xconnect, what happens when xconnect is specified in a playbook task and run against ND 3.x?

I know adding unit tests would be difficult due to the way UT are structured for the dcnm_network module, so I'm fine with not adding to or modifying unit tests.

I will leave it to @mikewiebe to approve.

Thank you!

plugins/modules/dcnm_network.py
```diff
@@ -319,6 +319,12 @@
         - Not applicable at Multisite parent fabric level
         type: str
         required: false
+      xconnect:
```
mikewiebe, line 322: Hey @ccoueffe, @mtarking and I were discussing that we don't want to document any parameters that showed up on legacy APIs because they were added to the new APIs.  Is this one of those?  If so, we need to remove this from the module docs.

mikewiebe:
### Medium: Enforce and document Xconnect's actual applicability contract

**Issue**

The new option accepts `xconnect: true` while `is_l2only` remains at its default `false`, even though Nexus Dashboard supports Xconnect only for Layer-2-only networks. The documentation also says `ND version > 4.1`, while the implementation enables NDFC `12.4.1` itself, and it does not explain that the option is omitted from the module's parent/child fabric schemas.

**Example 1 — invalid Layer-3 Xconnect is accepted**

```yaml
- cisco.dcnm.dcnm_network:
    fabric: Fabric-1
    state: merged
    config:
      - net_name: Xconnect-Network
        vrf_name: Tenant-1
        is_l2only: false
        gw_ip_subnet: 192.0.2.1/24
        xconnect: true
```

The module accepts this as:

```json
{
  "invalid": [],
  "xconnect": true,
  "is_l2only": false
}
```

and can construct:

```json
{
  "isLayer2Only": false,
  "xconnect": true
}
```

That combination is unsupported. The module should fail before contacting NDFC with an error such as `xconnect=true requires is_l2only=true`. Omitting `is_l2only` has the same problem because its default is `false`. A valid task must instead use:

```yaml
config:
  - net_name: Xconnect-Network
    is_l2only: true
    xconnect: true
```

**Example 2 — the documented version boundary disagrees with runtime**

The documentation says the feature is available only for `ND version > 4.1`, which excludes ND 4.1 itself. The code checks `_ndfc_version_gte("12.4.1")`, so it enables the option on NDFC `12.4.1`.

| Environment | Documentation implies | Runtime does |
| --- | --- | --- |
| ND 4.1 / NDFC 12.4.1 | Unsupported because it is not greater than 4.1 | Adds and accepts `xconnect` |
| Later than ND 4.1 | Supported | Adds and accepts `xconnect` |

If `12.4.1` is supported, the public contract must describe the boundary as inclusive.

**Example 3 — standalone accepts the option while parent fabrics reject it**

The valid Layer-2 task above is accepted for a standalone fabric. For a multisite or multicluster parent, `xconnect` is absent from the validation schema, so the same input produces an error equivalent to:

```text
Invalid parameters in playbook: xconnect
```

It is also absent from child-fabric schemas. That restriction may be intentional, but the documentation does not tell users whether Xconnect is standalone-only, accidentally missing from parent support, or expected at another fabric scope.

**Evidence**

- [`plugins/modules/dcnm_network.py` line 322](https://github.com/CiscoDevNet/ansible-dcnm/pull/612/files#diff-81913290492fc2a4c62ca8ade79047040d6afc496e3dbc92761e14a1ed6f472cR322) documents only a version restriction and says `> 4.1`.
- [`plugins/modules/dcnm_network.py` line 5229](https://github.com/CiscoDevNet/ansible-dcnm/pull/612/files#diff-81913290492fc2a4c62ca8ade79047040d6afc496e3dbc92761e14a1ed6f472cR5229) adds `xconnect` only to the standalone branch when NDFC is at least `12.4.1`; no validation ties it to `is_l2only`.
- A focused validation probe against the PR head returned `{"invalid": [], "xconnect": true, "is_l2only": false}`. A second probe showed the option present for `standalone` but absent for `multisite_parent`, `multicluster_parent`, `multisite_child`, and `multicluster_child`; parent validation reports `xconnect` as an extra parameter.
- Cisco's ND 4.1.1 configuration guide states that [VXLAN Xconnect supports only Layer-2 networks](https://www.cisco.com/c/en/us/td/docs/dcn/nd/4x/articles-411/working-with-fabric-segmentation-and-security-vxlan.html#enable-xconnect-for-layer-2-only-network) and documents an additional dot1q-attachment restriction.

**Existing PR overlap**

Related. [Allen's earlier review](https://github.com/CiscoDevNet/ansible-dcnm/pull/612#pullrequestreview-3657130719) asks for integration coverage and asks about older controllers, while [the active inline thread](https://github.com/CiscoDevNet/ansible-dcnm/pull/612#discussion_r3647629278) questions whether the option should be documented at all. Neither identifies the accepted Layer-3 configuration or the current standalone/parent contract mismatch.

**Impact**

A documented task can pass module validation and then fail late at the controller (or request an unsupported combination), while ND 4.1 and multi-fabric users receive contradictory guidance about whether the option is available. The definite runtime defect is accepting `xconnect: true` with `is_l2only: false`; the version and fabric-scope behaviors are contract inconsistencies that must be made explicit.

**Suggested fix**

Reject `xconnect: true` unless `is_l2only: true`, document the Layer-2 and attachment restrictions, make the version wording match the actual `12.4.1`/ND 4.1 threshold, and explicitly document the supported fabric scopes. The resulting contract should state that Xconnect requires `is_l2only: true`, identify the inclusive minimum controller release, list the supported fabric scopes, and include the controller's attachment restrictions.

Add the three examples above as focused tests for valid Layer-2 input, invalid Layer-3/defaulted input, version boundaries, parent/child behavior, payload construction, query, and idempotent merged/replaced updates.

mikewiebe:
### Medium: Do not silently turn a version lookup failure into an invalid option

**Issue**

The new full-version lookup catches every exception and returns `None`.
On a controller that supports Xconnect, a transient failure or routing
difference on this additional endpoint silently removes `xconnect` from the
validation schema instead of reporting the version lookup problem.

**Example — a supported ND 4.1 controller reports a misleading validation error**

Assume the controller is ND 4.1/NDFC 12.4.1, which supports Xconnect, and the
user supplies this valid Layer-2 task:

```yaml
- name: Create an Xconnect network
  cisco.dcnm.dcnm_network:
    fabric: FABRIC-1
    state: merged
    config:
      - net_name: Xconnect-Network
        is_l2only: true
        xconnect: true
```

If the module's additional exact-version request temporarily returns HTTP
503, times out, or produces an unexpected response, the broad exception
handler converts that failure to `None`. The version check then treats the
controller as older than NDFC 12.4.1, builds the validation schema without
`xconnect`, and rejects the task with:

```text
Invalid parameters in playbook: xconnect
```

That message incorrectly suggests that `xconnect` is misspelled or
unsupported. The actual problem is that the module could not determine the
controller version. The same task can therefore fail once and succeed
unchanged on a retry when the version request recovers. The module should
instead preserve and report an actionable error such as:

```text
Unable to determine the controller version required to validate xconnect:
version request returned HTTP 503.
```

**Evidence**

- [`plugins/modules/dcnm_network.py` line 5053](https://github.com/CiscoDevNet/ansible-dcnm/pull/612/files#diff-81913290492fc2a4c62ca8ade79047040d6afc496e3dbc92761e14a1ed6f472cR5053)
  issues an additional controller request, catches `Exception`, and returns
  `None` on any failure.
- [`plugins/modules/dcnm_network.py` line 5071](https://github.com/CiscoDevNet/ansible-dcnm/pull/612/files#diff-81913290492fc2a4c62ca8ade79047040d6afc496e3dbc92761e14a1ed6f472cR5071)
  treats a missing version as simply below the feature threshold.
- [`plugins/modules/dcnm_network.py` line 5229](https://github.com/CiscoDevNet/ansible-dcnm/pull/612/files#diff-81913290492fc2a4c62ca8ade79047040d6afc496e3dbc92761e14a1ed6f472cR5229)
  consequently omits `xconnect` from the schema, after which normal
  extra-parameter validation reports only `xconnect` as invalid.
- The existing 59-test network suite passes without exercising a successful
  full-version lookup or any Xconnect behavior, so this fallback is not
  distinguished from an intentionally unsupported controller.

**Existing PR overlap**

**Impact**

An otherwise valid task against ND 4.1 or later can fail with a misleading
`Invalid parameters in playbook: xconnect` message. The added request also
runs for every network operation, even when the task does not use Xconnect.

**Suggested fix**

Retrieve the exact controller version once in the action plugin and pass it in
`_fabric_details`, or otherwise preserve an explicit success/failure result.
If `xconnect` is requested and the version cannot be determined, fail with the
underlying actionable version error; avoid the extra lookup when it is not
needed. Cover success, unsupported-version, malformed-response, and transport
failure paths in action-plugin-aware unit tests.

mikewiebe (changes requested):
Requesting changes for the two unresolved functional findings:

- [Enforce and document Xconnect's actual applicability contract](https://github.com/CiscoDevNet/ansible-dcnm/pull/612#issuecomment-5079057835): the module currently accepts `xconnect: true` with `is_l2only: false`, and the supported version and fabric-scope contract is inconsistent.
- [Do not silently turn a version lookup failure into an invalid option](https://github.com/CiscoDevNet/ansible-dcnm/pull/612#issuecomment-5079081946): an exact-version lookup failure is converted to `None`, causing a valid Xconnect task to fail with the misleading `Invalid parameters in playbook: xconnect` error.

Please address these runtime and validation paths and add focused tests covering the supported configuration, invalid Layer-3 input, controller-version boundaries, and version-lookup failures.

mikewiebe:
### Medium: Handle explicit `xconnect: false` when controller version is indeterminate

**Issue**

Commit `10ecc104` partially addresses the existing version-lookup finding: `xconnect: true` now fails cleanly when the exact controller version cannot be determined.

However, the failure check is based on truthiness. An explicit `xconnect: false` bypasses it even though false is a real requested state.

Assume an NDFC 12.4.1 network currently has Xconnect enabled and the exact-version endpoint temporarily fails. The operator tries to disable it:

```yaml
- cisco.dcnm.dcnm_network:
    fabric: Fabric-1
    state: merged
    config:
      - net_name: Xconnect-Network
        is_l2only: true
        xconnect: false
```

The module does not reject this request. Because version discovery failed, both desired-payload and existing-state normalization omit `xconnect`. The transition from `true` to `false` disappears from the comparison, so the module can report no change without disabling Xconnect.

**Evidence**

- [`_get_ndfc_version()`](https://github.com/CiscoDevNet/ansible-dcnm/blob/10ecc104fcc22226fc45c9e3fb0b2b2f12f2f8dc/plugins/modules/dcnm_network.py#L5053-L5069) performs another version request, catches every exception, discards the underlying error, and returns `None`.
- [Input validation](https://github.com/CiscoDevNet/ansible-dcnm/blob/10ecc104fcc22226fc45c9e3fb0b2b2f12f2f8dc/plugins/modules/dcnm_network.py#L5388-L5403) checks lookup failure only inside `if net.get("xconnect", False) is True`; explicit false bypasses it.
- [Payload construction](https://github.com/CiscoDevNet/ansible-dcnm/blob/10ecc104fcc22226fc45c9e3fb0b2b2f12f2f8dc/plugins/modules/dcnm_network.py#L2492-L2497) and [existing-state normalization](https://github.com/CiscoDevNet/ansible-dcnm/blob/10ecc104fcc22226fc45c9e3fb0b2b2f12f2f8dc/plugins/modules/dcnm_network.py#L2595-L2600) both omit Xconnect when the version is unknown.
- The current 59-test network suite contains no Xconnect case.

**Existing PR overlap**

This is the remaining portion of the existing [version-lookup comment](https://github.com/CiscoDevNet/ansible-dcnm/pull/612#issuecomment-5079081946). The latest commit fixes its `xconnect: true` symptom but does not handle explicit false, query/readback, or merged-state preservation when the lookup is indeterminate.

**Impact**

During a transient version-endpoint failure, the module can report no change for an explicit request to disable Xconnect. It can also return incomplete query output or perform an unrelated update without preserving existing Xconnect intent.

**Suggested fix**

Gate validation on whether the original input contains the `xconnect` key, not on the value's truthiness. If the user supplies either boolean and the version is indeterminate, fail with the actionable lookup error.

Query and merged-state paths should also fail or preserve returned Xconnect state rather than silently removing it. Add focused coverage for true, false, omitted input, query, merged preservation, supported and unsupported versions, malformed responses, and transport failures.

mikewiebe (approved):

PR #663, opened 2026-04-07
Handle non-resolvable deferred interfaces in dcnm_interface

### LINKED ISSUE #658: dcnm_interface: overridden crashes when deferred non-deletable interfaces have null underlayPolicies
### Community Note

### Ansible Version and collection version

```text
ansible [core 2.16.3]
cisco.dcnm 3.11.0-dev
```

### DCNM version

- NDFC 12.5.0.475
- ND 4.2.1.10

Note: this environment is newer than the versions currently listed in the collection README, but the failure appears to be a null-safety issue in module logic rather than a usage problem.

### Affected module(s)

- `dcnm_interface`

### Ansible Playbook

The original failure was observed through the `cisco.nac_dc_vxlan` remove role, which calls `cisco.dcnm.dcnm_interface` with `state: overridden`. The failing call shape is:

```yaml
---
- hosts: ndfc
  gather_facts: no
  collections:
    - cisco.dcnm

  tasks:
    - name: Override interfaces in ISN fabric
      cisco.dcnm.dcnm_interface:
        fabric: nac-mcfg-isn
        state: overridden
        deploy: false
        config:
          - name: lo0
            type: lo
            switch:
              - 10.122.226.100
            profile:
              mode: lo
              admin_state: true
              description: "Loopback Interface 0"
              ipv4_addr: 10.101.101.101
              route_tag: "54321"
```

In this test, `lo0` is the only model-defined managed interface. The crash is not caused by the loopback payload itself; it happens later while reconciling all interfaces returned by NDFC for the switch.

### Debug Output

Full traceback obtained by running with `ANSIBLE_NETWORK_IMPORT_MODULES=0` to force subprocess execution:

```text
Traceback (most recent call last):
  File ".../AnsiballZ_dcnm_interface.py", line 107, in <module>
    _ansiballz_main()
  File ".../AnsiballZ_dcnm_interface.py", line 99, in _ansiballz_main
    invoke_module(zipped_mod, temp_path, ANSIBALLZ_PARAMS)
  File ".../AnsiballZ_dcnm_interface.py", line 47, in invoke_module
    runpy.run_module(mod_name='ansible_collections.cisco.dcnm.plugins.modules.dcnm_interface', ...)
  File "<frozen runpy>", line 226, in run_module
  File "<frozen runpy>", line 98, in _run_module_code
  File "<frozen runpy>", line 88, in _run_code
  File ".../dcnm_interface.py", line 6380, in <module>
  File ".../dcnm_interface.py", line 6343, in main
  File ".../dcnm_interface.py", line 5338, in dcnm_intf_get_diff_overridden
TypeError: 'NoneType' object is not subscriptable
```

The failing code at line 5338:

```python
for intf in defer_list:
    delem = {}
    sno = intf["serialNo"]
    fabric = intf["fabricName"]
    name = intf["underlayPolicies"][0]["source"]   # <-- crash: underlayPolicies is None
```

Default Ansible output (without `ANSIBLE_NETWORK_IMPORT_MODULES=0`):

```text
TASK [cisco.nac_dc_vxlan.remove : Remove Unmanaged Fabric Interfaces in Nexus Dashboard - Diff Run Feature Disabled]
fatal: [nac-mcfg-isn]: FAILED! => {
    "changed": false,
    "module_stderr": "'NoneType' object is not subscriptable",
    "module_stdout": "",
    "msg": "MODULE FAILURE\nSee stdout/stderr for the exact error"
}
```

### Expected Behavior

`state: overridden` should safely reconcile switch interfaces against the supplied config.

If NDFC returns deferred/non-deletable interfaces with `underlayPolicies: null`, the module should skip them or handle them defensively. It should not crash.

### Actual Behavior

The module crashes during diff calculation for `state: overridden`.

The failure occurs after the module fetches all interfaces for the switch and processes deferred interfaces, not while validating the input loopback config.

### Steps to Reproduce

1. Use an ISN fabric where the model-defined managed interface set includes a loopback such as `lo0`.
2. Invoke `cisco.dcnm.dcnm_interface` with `state: overridden`.
3. The module queries NDFC for all interfaces on the switch.
4. In our NDFC 12.5.0.475 / ND 4.2.1.10 environment, many physical Ethernet interfaces are returned with:
   - `deletable: false`
   - `underlayPolicies: null`
5. Those interfaces are added to the deferred list.
6. During deferred processing, the module assumes `underlayPolicies[0]["source"]` exists and crashes with:
   - `'NoneType' object is not subscriptable`

### Additional Observations

The behavior differs between our older ND 4.1 ISN environment and NDFC 12.5 / ND 4.2:

| Observation | Older ND 4.1 ISN env | NDFC 12.5.0.475 / ND 4.2.1.10 |
| --- | --- | --- |
| Interfaces with `underlayPolicies: null` | 64 | 65 |
| Interfaces with `deletable: false` | 0 | 65 |

The important part is that `underlayPolicies: null` already existed before.

What appears to have changed is the `deletable` flag:

- In the older environment, those Ethernet interfaces were `deletable: true`, so they were processed directly and never reached the deferred path.
- In NDFC 12.5 / ND 4.2, the same class of interfaces is returned as `deletable: false`, so they are added to the deferred list and later hit the unsafe dereference.

So the crash trigger appears to be this combination:

- deferred / non-deletable interface
- `underlayPolicies: null`

rather than the loopback config itself.

### Suspected Root Cause

In the overridden diff path, non-deletable interfaces are added to a deferred list first.

Later, deferred processing assumes `underlayPolicies[0]["source"]` is always present. When NDFC returns `underlayPolicies: null`, that assumption is invalid and the module crashes.

### Suggested Direction

A defensive fix in the deferred path would likely solve this class of failure:

- guard against `underlayPolicies` being `null` or empty before dereferencing
- skip or safely handle deferred interfaces without a parent/source relationship
- add a unit test covering:
  - `state: overridden`
  - deferred interface
  - `deletable=false`
  - `underlayPolicies=null`

### References

This issue was first observed through `cisco.nac_dc_vxlan`, but the failing logic appears to be in the base `cisco.dcnm.dcnm_interface` module.

dacasti2mx:
Fixes #658

## Summary

This change fixes a crash in `dcnm_interface` during `state=overridden` when the module processes non-deletable Ethernet interfaces returned by NDFC without a resolvable underlay policy source.

## Problem

In the deferred override flow, the module assumes that every non-deletable Ethernet interface has:

```python
underlayPolicies[0]["source"]
```

That is not always true.

With NDFC 4.2, ISN and External fabrics can return physical Ethernet interfaces with:

- `deletable: false`
- `underlayPolicies: null`

When those interfaces enter the deferred path, the module crashes with:

```text
TypeError: 'NoneType' object is not subscriptable
```

## Root Cause

The deferred flow treats all `deletable=false` Ethernet interfaces as if they were resolvable later via `underlayPolicies[].source`.

That assumption is invalid for protected or unmanaged Ethernet interfaces in ISN and External fabrics.

## Changes

- Added a helper to safely extract a usable underlay policy `source`
- Added a helper to record non-resolvable deferred interfaces under `diff.skipped`
- Updated the override flow so that Ethernet interfaces are only added to `deferred` when:
  - `deletable == false`, and
  - a valid `underlayPolicies[].source` exists
- Added a defensive check in the deferred loop to avoid direct indexing of `underlayPolicies[0]["source"]`
- Added a unit test that covers:
  - `state=overridden`
  - physical Ethernet
  - `deletable=false`
  - `underlayPolicies=None`

## Behavior After This Change

For non-deletable Ethernet interfaces without a resolvable policy source, the module now:

- does not crash
- does not add them to `deferred`
- reports them in `diff.skipped`
- leaves them untouched

This preserves the existing deferred behavior for interfaces that do have a valid `source`.

## Validation

Manual validation:

- NDFC 4.2
- ISN fabric: remove flow completed successfully
- External fabric: remove flow completed successfully
- Problematic physical Ethernet interfaces were reported under `skipped`
- No `NoneType` or `subscriptable` failure was observed

Unit validation:

- Added coverage for `deletable=false` plus `underlayPolicies=None` in `test_dcnm_intf.py`

### DIFF
--- a/tests/unit/modules/dcnm/test_dcnm_intf.py
+++ b/tests/unit/modules/dcnm/test_dcnm_intf.py
@@ -17,6 +17,7 @@ from __future__ import absolute_import, division, print_function
 
 __metaclass__ = type
 
+import copy
 from unittest.mock import patch
 
 # from units.compat.mock import patch
@@ -194,6 +195,63 @@ class TestDcnmIntfModule(TestDcnmModule):
                 self.playbook_mock_succ_resp,
                 self.playbook_mock_succ_resp,
             ]
+
+        if (
+            "test_dcnm_intf_override_eth_intf_types_skip_non_resolvable_deferred"
+            in self._testMethodName
+        ):
+
+            playbook_have_all_data = copy.deepcopy(
+                self.have_all_payloads_data.get("payloads")
+            )
+            for intf in playbook_have_all_data["DATA"]:
+                if intf["ifName"] == "Ethernet1/1":
+                    intf["deletable"] = "False"
+                    intf["underlayPolicies"] = None
+                    break
+
+            eth_1_2_access_intf = self.have_all_payloads_data.get(
+                "eth_1_2_access_payload"
+            )
+            eth_3_2_access_intf = self.have_all_payloads_data.get(
+                "eth_3_2_access_payload"
+            )
+            self.breakout_policies_data = loadPlaybookData(
+                "dcnm_intf_breakout_policies"
+            )
+            empty_breakout_resp = self.breakout_policies_data.get(
+                "empty_breakout_policies"
+            )
+
+            def dcnm_send_side_effect(*args, **kwargs):
+                path = args[2]
+
+                if path.endswith("/accessmode"):
+                    return self.mock_monitor_false_resp
+                if "/control/policies/switches/" in path:
+                    return empty_breakout_resp
+                if "interface/detail?serialNumber=" in path:
+                    return playbook_have_all_data
+                if (
+                    "interface?serialNumber=" in path
+                    and "ifName=Ethernet1/1" in path
+                ):
+                    raise AssertionError(
+                        "Skipped deferred interface should not be queried again"
+                    )
+                if (
+                    "interface?serialNumber=" in path
+                    and "ifName=Ethernet1/2" in path
+                ):
+                    return eth_1_2_access_intf
+                if (
+                    "interface?serialNumber=" in path
+                    and "ifName=Ethernet3/2" in path
+                ):
+                    return eth_3_2_access_intf
+                return self.playbook_mock_succ_resp
+
+            self.run_dcnm_send.side_effect = dcnm_send_side_effect
 
         if (
             "test_dcnm_intf_override_eth_intf_types_only"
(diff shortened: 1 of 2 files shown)


dacasti2mx:
Hi @mikewiebe, when you have time, could you please provide feedback for this PR?
thanks.

mikewiebe:
## PR 663 Integration Validation

Validated PR 663 against a real Nexus Dashboard / NDFC controller.

### Environment

- ND controller: `10.122.84.112`
- Fabric: `nac-msd-fabric1`
- Switch tested: `10.122.84.185`
- Switch serial: `9FV9WJOZ0PE`
- PR head tested: `ccb9515f9270d301228e8da4f0a77620caca1a8d`

### Test Scope

Ran a targeted integration validation for the code path changed by this PR:

- Module: `cisco.dcnm.dcnm_interface`
- State: `overridden`
- `override_intf_types: [eth]`
- `deploy: false`
- Ansible check mode: enabled

This exercises the PR-specific behavior for non-deletable Ethernet interfaces without a resolvable `underlayPolicies.source`, without applying changes to the controller or switch.

### Results

- Play recap: `ok=6`, `changed=0`, `failed=0`, `unreachable=0`
- Target Ethernet interfaces read from NDFC: `66`
- Target non-deletable Ethernet interfaces: `64`
- PR-specific skipped condition observed: `404`
- Deferred count: `0`
- Deleted count: `0`
- Replaced count: `0`
- Total skipped count: `448`

### Conclusion

The PR-specific override/deferred logic passed validation on a real ND/NDFC controller. The module completed the Ethernet `state=overridden` diff path successfully, skipped non-resolvable deferred interfaces as expected, and did not mutate the controller or switch state during the check-mode run.

tests/unit/modules/dcnm/test_dcnm_intf.py
```diff
@@ -6078,6 +6151,65 @@ def test_dcnm_intf_override_eth_intf_types_only(self):
...
+                override_intf_types=["eth"],
+                deploy=False,
+                config=[],
+            )
+        )
+        result = self.execute_module(changed=True, failed=False)
+
+        self.assertEqual(len(result["diff"][0]["deferred"]), 0)
+        self.assertEqual(len(result["diff"][0]["deleted"]), 0)
+        self.assertLessEqual(len(result["diff"][0]["replaced"]), 2)
```
mikewiebe: **Low:** `tests/unit/modules/dcnm/test_dcnm_intf.py:6188` uses:

```python
self.assertLessEqual(len(result["diff"][0]["replaced"]), 2)
```

This would still pass if zero or one valid Ethernet interfaces were replaced. Since this test is meant to prove only Ethernet1/1 is skipped while Ethernet1/2 and Ethernet3/2 still flow normally, I’d make the assertion exact:

```python
self.assertEqual(len(result["diff"][0]["replaced"]), 2)
self.assertEqual(
    {
        intf["interfaces"][0]["ifName"].lower()
        for intf in result["diff"][0]["replaced"]
    },
    {"ethernet1/2", "ethernet3/2"},
)
```

mikewiebe (changes requested):
@dacasti2mx Minor comment in the test but otherwise looks good

dacasti2mx:
Thanks Mike. Fixed.

mikewiebe (approved):

PR #665, opened 2026-04-27
Bug fix for l2only network comparisons in an MCFG fabric

### LINKED ISSUE #668: Fix for l2only network comparisons in an MCFG fabric
### Community Note

When you want to make an update to a layer2 only network under an MCFG fabric, the module would try and use the create new network workflow rather than the update existing network workflow due a slight difference in the response from NDFC for these networks. Under an MSD or standalone fabric, the networkTemplateConfig dictionary contains a vrfName key which was used to determine whether an l2only network exists or not, under an MCFG fabric this key doesn't exist, so we added support to fall back on a VRF key in the root of the response if the vrfName key doesn't exist within networkTemplateConfig.

### DCNM version

* All

### Affected module(s)

* dcnm_network

### Ansible Playbook

```yaml
# Copy-paste your anisble playbook here 
```

mthurstocisco:
Resolves https://github.com/CiscoDevNet/ansible-dcnm/issues/668

### DIFF
--- a/plugins/module_utils/network/dcnm/dcnm.py
+++ b/plugins/module_utils/network/dcnm/dcnm.py
@@ -37,7 +37,8 @@ except ImportError:
 FEDERATION_MANAGER_NOT_FOUND_ERRORS = [
     'Invalid JSON response: this API is allowed only for remote user',
     'A federation manager does not exist',
-    'Invalid JSON response: cannot serve APIs as federation state is secondary. Use primary cluster for APIs'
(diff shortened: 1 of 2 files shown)


mthurstocisco:
Just to confirm, the previous check looked in the nested dictionary for the vrfName key, when you create a l2only network in an MCFG fabric, using the module, you don't get that key generated so the fix adds a backup look up for the vrf key in the top layer dictionary therefore not touching the original logic for MSD or standalone networks, just adding a fall back in the case of an MCFG network created using the module

mikewiebe (approved):

PR #671, opened 2026-05-04
Additional VRF pfmc Improvements

mikewiebe:
## Summary

This PR optimizes the `dcnm_vrf` module for large-scale no-change and targeted VRF runs.

The main improvement is reducing local Python CPU time in the VRF diff path. In the measured no-change full sync, the VRF task was not waiting on NDFC API calls; it was spending most of its time doing local attachment comparison, repeated config scans, hot-path stack inspection, and eager debug serialization.

### Faster VRF Attachment Diffing

- Optimized `diff_for_attach_deploy()` by indexing existing attachments by `(serialNumber, vrfName)`.
- Replaced nested wanted-vs-have attachment scans with direct dictionary lookups.
- Preserved existing comparison behavior for:
  - attachment state
  - deployment state
  - instance values
  - VRF-Lite extension values
  - replace-mode behavior

### Cached Config Lookups

- Added cached config indexes:
  - `config_by_vrf_name`
  - `vrf_lite_vrf_names`
  - `any_vrf_lite_in_config`
- Added `_refresh_config_indexes()` and refresh it after validation.
- Replaced repeated linear searches through `self.config` with O(1) lookups.
- Used cached wanted VRF names in targeted `get_have()` filtering.

### Faster Replace Path

- Added `want_attach_by_name` for O(1) wanted attachment lookup.
- Replaced nested replace-mode attachment comparison with per-VRF serial-number sets.
- Maintained index updates in rollback and child-fabric paths.

### Reduced Hot-Path Overhead

- Removed expensive `inspect.stack()` calls from high-frequency helper paths:
  - `to_bool()`
  - `update_attach_params()`
  - `ip_to_serial_number()`
- Guarded large debug-only serialization behind `log.isEnabledFor(logging.DEBUG)`.
- This avoids building large `json.dumps(..., indent=4)` strings during normal runs.

### VRF-Lite Detection Optimization

- Replaced repeated full-config scans in `_has_vrf_lite_in_config()` with cached set membership.
- This avoids repeated scanning for every VRF and attachment when VRF-Lite is not in use.

### Delete-Scale Safety

- Chunked VRF attachment readiness polling in `wait_for_vrf_attachments_del_ready()` into batches of 30.
- Chunked bulk VRF delete calls into batches of 30.
- This keeps query strings bounded for large delete operations.

## Intentionally Deferred

- A VRF-specific bulk-update endpoint probe was tested but not included in this PR.
- The current shared bulk probe behavior remains unchanged because the VRF unit tests rely on existing mocked `dcnm_send()` call ordering.
- A future PR can add entity-specific bulk probing with corresponding unit test fixture updates.

## Validation

Unit tests:

    source env.sh ; pytest collections/ansible_collections/cisco/dcnm/tests/unit/modules/dcnm/test_dcnm_vrf*

Result:

    63 passed in 6.08s

Additional checks:

    python3 -m py_compile collections/ansible_collections/cisco/dcnm/plugins/modules/dcnm_vrf.py
    git diff --check

Both passed.

Full playbook verification:

    source env.sh ; ansible-playbook -i inventory.yaml vxlan.yaml

Result:

    Playbook run took: 0:54:03
    Recap: ok=48 changed=4 unreachable=0 failed=0 skipped=18
    Deploy: changed=0/186, no switches needed deployment

## Performance Results

Measured against the previous no-change full sync baseline.

| Metric | Before | After | Improvement |
|---|---:|---:|---:|
| CREATE vrfs | 987.4s | 175.6s | -811.8s / ~82% faster |
| Create pipeline | 2032.6s | 1157.9s | -874.7s |
| Full playbook | 1:09:07 | 0:54:03 | -15:04 |

Networks remained stable, which confirms the change is isolated to VRF behavior:

| Metric | Before | After |
|---|---:|---:|
| CREATE networks | 301.1s | 304.4s |

## Functional Impact

- No fabric resource changes were detected during the verification run.
- `CREATE vrfs` completed with `changed=False`.
- Deploy found `0/186` switches needing deployment.
- Existing VRF unit tests all pass.
- Existing bulk API support behavior remains unchanged.

### DIFF
--- a/plugins/modules/dcnm_vrf.py
+++ b/plugins/modules/dcnm_vrf.py
@@ -1076,18 +1076,24 @@ class DcnmVrf:
         self.params = module.params
         self.state = self.params.get("state")
 
-        msg = f"self.state: {self.state}, "
-        msg += "self.params: "
-        msg += f"{json.dumps(self.params, indent=4, sort_keys=True)}"
-        self.log.debug(msg)
+        if self.log.isEnabledFor(logging.DEBUG):
+            msg = f"self.state: {self.state}, "
+            msg += "self.params: "
+            msg += f"{json.dumps(self.params, indent=4, sort_keys=True)}"
+            self.log.debug(msg)
 
         self.fabric = module.params["fabric"]
         self.config = copy.deepcopy(module.params.get("config"))
+        self.config_by_vrf_name = {}
+        self.vrf_lite_vrf_names = set()
+        self.any_vrf_lite_in_config = False
+        self._refresh_config_indexes()
 
-        msg = f"self.state: {self.state}, "
-        msg += "self.config: "
-        msg += f"{json.dumps(self.config, indent=4, sort_keys=True)}"
-        self.log.debug(msg)
+        if self.log.isEnabledFor(logging.DEBUG):
+            msg = f"self.state: {self.state}, "
+            msg += "self.config: "
+            msg += f"{json.dumps(self.config, indent=4, sort_keys=True)}"
+            self.log.debug(msg)
 
         # Setting self.conf_changed to class scope since, after refactoring,
         # it is initialized and updated in one refactored method
@@ -1119,6 +1125,7 @@ class DcnmVrf:
         # O(1) lookup dict populated by get_have()
         self.have_attach_by_name = {}
         self.want_attach = []
+        self.want_attach_by_name = {}
         self.diff_attach = []
         self.validated = []
         # diff_detach contains all attachments of a vrf being deleted,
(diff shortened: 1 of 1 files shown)


plugins/modules/dcnm_vrf.py
```diff
@@ -1076,18 +1076,24 @@ def __init__(self, module):
         self.params = module.params
         self.state = self.params.get("state")
 
-        msg = f"self.state: {self.state}, "
-        msg += "self.params: "
-        msg += f"{json.dumps(self.params, indent=4, sort_keys=True)}"
-        self.log.debug(msg)
+        if self.log.isEnabledFor(logging.DEBUG):
```
mikewiebe, line 1079: Need to check these self.log.isEnabledFor(logging.DEBUG) to see why codex thought this was a good idea and if this is really needed

mikewiebe: Found the answer

Guarded large debug-only serialization behind log.isEnabledFor(logging.DEBUG) 

But.. we should not be logging anyway if the logging json file and env var is not enabled so not sure this is really needed

PR #685, opened 2026-06-09
Update dcnm_vpc_pair_utils.py

ccoueffe:
add condition when switch is in not in-sync like preprovision.

### DIFF
--- a/plugins/module_utils/network/dcnm/dcnm_vpc_pair_utils.py
+++ b/plugins/module_utils/network/dcnm/dcnm_vpc_pair_utils.py
@@ -766,12 +766,14 @@ def dcnm_vpc_pair_utils_process_deploy_payloads(self, deploy_list):
     resp = None
     deploy_flag = False
 
-    if deploy_list:
-        # Perform a config-save first before config-deploy
-        dcnm_vpc_pair_utils_save_config_changes(self)
-        dcnm_vpc_pair_utils_invalidate_sync_cache(self)
-    else:
+    if not deploy_list:
         return deploy_flag
+
+    if not self.managable:
+        return deploy_flag
+
+    dcnm_vpc_pair_utils_save_config_changes(self)
+    dcnm_vpc_pair_utils_invalidate_sync_cache(self)
 
     for elem in deploy_list:
         rc, resp = dcnm_vpc_pair_utils_deploy_elem(self, elem)
--- a/plugins/module_utils/network/dcnm/dcnm_vpc_pair_utils.py
+++ b/plugins/module_utils/network/dcnm/dcnm_vpc_pair_utils.py
@@ -658,7 +658,7 @@ def dcnm_vpc_pair_utils_get_sync_status(self, elem):
     switches = [
         self.sn_ip[elem[peer]]
         for peer in ["peerOneId", "peerTwoId"]
-        if self.sn_ip.get(elem[peer], "") in self.managable
+        if not self.managable or self.sn_ip.get(elem[peer], "") in self.managable
     ]
 
     if not switches:
@@ -731,7 +731,7 @@ def dcnm_vpc_pair_utils_deploy_elem(self, elem):
     for peer in ["peerOneId", "peerTwoId"]:
         # Skip deploy for switches that are not managable (e.g. preprovision/unreachable)
         peer_ip = self.sn_ip.get(elem[peer], "")
-        if peer_ip not in self.managable:
+        if self.managable and peer_ip not in self.managable:
             continue
 
         path = self.paths["VPC_PAIR_DEPLOY_PATH"].format(elem["fabric"], elem[peer])
--- a/plugins/module_utils/network/dcnm/dcnm_vpc_pair_utils.py
+++ b/plugins/module_utils/network/dcnm/dcnm_vpc_pair_utils.py
@@ -733,7 +733,7 @@ def dcnm_vpc_pair_utils_deploy_elem(self, elem):
         peer_ip = self.sn_ip.get(elem[peer], "")
         if peer_ip not in self.managable:
             continue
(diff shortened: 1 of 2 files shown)


dacasti2mx:
## 2 scenarios tested

  Verified live with `epdb` in two scenarios.

  ---

  ### Scenario 1 — Semi-deployed (the #594 target): **works**

  **Conditions**
  - Spine / Leaf / BGW reachable
  - vPC members (Leaf-103 / Leaf-104) pre-provisioned, mgmt down

  **Evidence** — inside `dcnm_vpc_pair_utils_get_sync_status`:
  ```
  self.managable = {                          # the reachable switches
      '192.168.1.101': …, '192.168.1.108': …, '192.168.1.110': …,
      'Leaf-101': …, 'Spine-1201': …, 'BGW-1': …,
  }
  '192.168.1.103' in self.managable   →  False     # member not manageable
  switches                            →  []         # both members filtered out
  ```

  **Result:** `if not switches: return "In-Sync"` fires → the `overridden` remove finishes
  cleanly (`changed: false`, no *"did not reach In-Sync"*). **The fix works as intended.** 🙂

  ---

  ### Scenario 2 — All switches pre-provisioned: **does not work**

  **Conditions**
  - Every switch pre-provisioned, mgmt down

  **Evidence**
  ```
  self.managable                      →  {}          # nobody manageable
  inventory "managable" flags         →  all False
  '192.168.1.103' in self.managable   →  False
  switches                            →  ['192.168.1.103', '192.168.1.104']   # BOTH kept
  ```

  **Why:** with `self.managable` empty, `not self.managable` is `True`, so **both** peers
  are kept → `switches` is non-empty → `return "In-Sync"` never fires → **the skip does
  not engage.**

  The remove then fails at config-save with **HTTP 500**:
  ```
  "Switch(es) are either reloading or not reachable…"        # virtual peer-link
  "Member port Ethernet1/3 does not exist on the switch"     # physical peer-link
  ```

  → **All-preprov remove isn't handled by this PR** (skip inert + config-save 500).

  ---

  ### Note from reading the code (not tested yet)

  `if not switches: return "In-Sync"` reports In-Sync **without querying NDFC**. Worth
  confirming this can't hide a real out-of-sync on a switch that's only briefly unreachable.

  ---

  ### Verdict

  **Approved for the semi-deployed scenario it targets** (verified live).
  The **all-unreachable** case is a known limitation — worth a note or a follow-up.

plugins/module_utils/network/dcnm/dcnm_vpc_pair_utils.py
```diff
@@ -655,7 +655,14 @@ def dcnm_vpc_pair_utils_get_sync_status(self, elem):
         'In-sync', if the switches configuration is "In-Sync" state, 'Not-In-Sync' otherwise
     """
 
-    switches = [self.sn_ip[elem["peerOneId"]], self.sn_ip[elem["peerTwoId"]]]
+    switches = [
+        self.sn_ip[elem[peer]]
+        for peer in ["peerOneId", "peerTwoId"]
+        if not self.managable or self.sn_ip.get(elem[peer], "") in self.managable
```
dacasti2mx, line 661: I will replace the entire  if statement  with this one:

if str(
            self.inventory_data.get(self.sn_ip.get(elem[peer], ""), {}).get("managable", "true")

previous logic:
include this vPC peer if the self.managable is empty or if its IP in inside self.managable
this will break the use case where we have a full pre-prov fabric where self.managable=[]

new logic:
Include this vPC peer if NDFC identifies this indivicual member as managable
This will covers the full pre-prov case

dacasti2mx (changes requested):
I will replace the entire  if statement  with this one:

if str(
            self.inventory_data.get(self.sn_ip.get(elem[peer], ""), {}).get("managable", "true")

previous logic:
include this vPC peer if the self.managable is empty or if its IP in inside self.managable
this will break the use case where we have a full pre-prov fabric where self.managable=[]

new logic:
Include this vPC peer if NDFC identifies this indivicual member as managable
This will covers the full pre-prov case

dacasti2mx (approved):
Tested `20c4fb92` in my lab (ND 4.1.1g / NDFC 12.4.1.321) — **the new guard fixes the all-unreachable case, for both create and remove.** 

Setup: VXLAN_EVPN fabric, all 5 switches pre-provisioned with mgmt down → `self.managable == {}`, vPC pair with virtual peer-link. This was the corner case where the previous revision still died on the module-internal config-save (HTTP 500 "Switch(es) are either reloading or not reachable").

Debugger placed right on the new guard, live run:

```
-> if not self.managable:
(Pdb) p self.managable
{}                                          <-- no manageable device
(Pdb) p deploy_list
[{'fabric': 'PREPROV3', 'peerOneId': 'PRELEAF103', 'peerTwoId': 'PRELEAF104'}]
                                            <-- a real deploy WAS pending
(Pdb) n
-> return deploy_flag                       <-- early return: config-save never runs
```

Remove (the netascode/nac-vxlan#594 scenario): with the pair dropped from the desired state, the DELETE goes through, the guard skips its deploy, and the task ends `ok (changed=True)`, playbook `failed=0`. Confirmed over REST that the pair is really gone from NDFC — decommission completes instead of aborting halfway. The partial-fabric case was already validated in the earlier round and is untouched by this revision (guard only triggers on empty `managable`).

One non-blocking note: in this state `get_sync_status` is still inert (the `not self.managable` fallback includes everyone), so every run queues a reconciliation deploy that the guard then discards. A per-member `managable` check from `inventory_data` would report In-Sync/no-op at the root — fine as a follow-up.

juburnet:
@mikewiebe - this is ready for final review and merge

mikewiebe (approved):

PR #710, opened 2026-07-25
Fix physical Ethernet state=deleted reset on ND 4.2

### LINKED ISSUE #709: dcnm_interface: state=deleted silently skips physical Ethernet reset on ND 4.2
### Community Note

* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize this request

### Ansible Version and collection version

```text
ansible [core 2.16.3]
cisco.dcnm 3.12.1-dev
```

The behavior was reproduced with the current `develop` implementation and the
same module revision was used for the ND 4.1 versus ND 4.2 comparison.

### DCNM version

* ND 4.1 / NDFC 12.4.1.321: working baseline
* ND 4.2 / NDFC 12.5.0.475: affected

### Affected module(s)

* `dcnm_interface`

### Ansible Playbook

```yaml
---
- name: Reset a physical Ethernet interface to its role default
  cisco.dcnm.dcnm_interface:
    fabric: test_fabric
    state: deleted
    deploy: false
    config:
      - name: Ethernet1/49
        type: eth
        switch:
          - 192.0.2.10
```

The interface has an existing nondefault description before this task runs.
The management address above is an anonymized example.

### Debug Output

The relevant sanitized capability and module results are:

```text
ND 4.1 / NDFC 12.4.1.321:
  deletable: true
  editAllowed: true
  first state=deleted run:
    changed: true
    replaced: 1 physical Ethernet role-default payload
  identical repeat:
    changed: false

ND 4.2 / NDFC 12.5.0.475:
  deletable: false
  editAllowed: true
  state=deleted check mode and real run:
    changed: false
    replaced: []
    deploy: []
```

Complete private-lab logs contain controller addresses and authentication
material, so they are not attached publicly. Sanitized request/response
details can be supplied if needed.

### Expected Behavior

Named `state: deleted` for a physical Ethernet interface should reset the
interface to the role-default policy, as it does with NDFC 12.4.1.321.

The operation should:

* construct one role-default replacement when the interface differs from its
  role default;
* clear the managed nondefault intent;
* report `changed: true` on the first run; and
* be idempotent on an identical repeat.

`deletable: false` describes whether the current policy can be deleted. It
should not prevent an editable physical interface from being updated to its
role-default policy when NDFC also reports `editAllowed: true`.

### Actual Behavior

On NDFC 12.5.0.475, named physical Ethernet deletion is a silent no-op.

The module reaches the physical-interface reset path but exits early when the
controller reports `deletable: false`. It never consults `editAllowed`, never
constructs the role-default replacement, reports `changed: false`, and leaves
the existing interface intent unchanged.

The same module code works with NDFC 12.4.1.321 because that controller returns
`deletable: true` for the same class of physical interfaces.

### Steps to Reproduce

1. Manage a physical Ethernet interface on ND 4.2 / NDFC 12.5.0.475.
2. Give the interface a harmless nondefault property such as a unique
   description.
3. Confirm that the interface API returns:

   ```text
   deletable: false
   editAllowed: true
   ```

4. Run the playbook above in check mode.
5. Observe an empty replacement diff and `changed: false`.
6. Run the task normally and confirm that the controller intent is unchanged.
7. Run the same code and test sequence on ND 4.1 / NDFC 12.4.1.321, where the
   controller returns `deletable: true`; observe one successful role-default
   replacement followed by an unchanged repeat.

### References

* Physical Ethernet reset support was introduced in #21. Its deleted-state
  fixtures use `deletable: true`.
* #658 documents a related ND 4.2 controller-metadata change affecting the
  overridden/deferred path. That issue fixed a null dereference; it does not
  address this named `state: deleted` silent no-op.
* The current workaround is an explicit `state: replaced` configuration
  containing the appropriate role-default interface settings.

### Suggested Fix

Preserve the existing NDFC 12.4 behavior and allow a physical Ethernet
interface to continue through the role-default replacement path when either:

* its current policy is deletable; or
* the physical interface is editable (`editAllowed: true`), even if the policy
  itself is not deletable.

Do not attempt to delete a non-deletable policy. Retain the existing
parent-member and underlay-source safeguards, explicitly report interfaces
that are neither deletable nor editable, and add regression coverage for both
controller capability combinations.

mikewiebe:
## Summary

ND 4.2 reports physical Ethernet interfaces as `deletable: false` and
`editAllowed: true`. `dcnm_interface` previously gated `state: deleted` only
on `deletable`, so a requested reset was silently skipped.

This change:

- accepts `editAllowed` as the ND 4.2 capability for resetting a physical
  Ethernet interface to its default policy;
- preserves the existing ND 4.1 `deletable` behavior;
- limits the compatibility fallback to `state: deleted`, leaving true
  `overridden` behavior unchanged;
- preserves port-channel/member and underlay-policy dependency handling;
- reports an explicit skipped result when neither capability permits the
  reset, or when an unresolved underlay dependency makes the fallback unsafe;
- adds coverage for named deletes, delete-all and switch-only forms, check
  mode, idempotency, string and boolean capability values, unsupported
  interfaces, and underlay dependencies.

`state: deleted` resets controller intent to the default Ethernet policy; it
does not delete the physical port.

## Validation

- ND 4.1 / NDFC 12.4.1 live baseline: check mode proposed one default
  replacement; the real run produced one replacement with no deployment; the
  repeat run was idempotent
- ND 4.2 / NDFC 12.5.0 pre-fix reproduction: `deletable: false`,
  `editAllowed: true`, and the module returned no diff
- ND 4.2 / NDFC 12.5.0 post-fix verification: named delete produced one
  default-policy replacement with no deployment; the repeat run was
  idempotent; the original controller intent was restored exactly after the
  test

Fixes #709

### DIFF
--- a/plugins/modules/dcnm_interface.py
+++ b/plugins/modules/dcnm_interface.py
@@ -5467,6 +5467,10 @@ class DcnmIntf:
 
         return None
 
+    def dcnm_intf_capability_enabled(self, intf, capability):
+
+        return str(intf.get(capability)).strip().lower() == "true"
+
     def dcnm_intf_skip_non_resolvable_deferred(self, intf):
 
         self.changed_dict[0]["skipped"].append(
@@ -5476,6 +5480,37 @@ class DcnmIntf:
                 "Deletable": intf.get("deletable"),
                 "Underlay Policies": intf.get("underlayPolicies"),
                 "Reason": "Non-deletable interface without resolvable underlay policy source",
+            }
+        )
+
+    def dcnm_intf_skip_physical_default_not_allowed(self, intf):
+
+        self.changed_dict[0]["skipped"].append(
+            {
+                "Name": intf["ifName"],
+                "Alias": intf.get("alias"),
+                "Deletable": intf.get("deletable"),
+                "Edit Allowed": intf.get("editAllowed"),
+                "Reason": (
+                    "Physical interface reset is not allowed because neither "
+                    "deletable nor editAllowed is true"
+                ),
+            }
+        )
+
+    def dcnm_intf_skip_edit_allowed_underlay_dependency(self, intf):
+
+        self.changed_dict[0]["skipped"].append(
+            {
+                "Name": intf["ifName"],
+                "Alias": intf.get("alias"),
+                "Deletable": intf.get("deletable"),
+                "Edit Allowed": intf.get("editAllowed"),
+                "Underlay Policies": intf.get("underlayPolicies"),
+                "Reason": (
+                    "Physical interface reset through editAllowed was skipped "
+                    "because its underlay policy source is not being deleted"
+                ),
             }
         )
 
@@ -5644,24 +5679,34 @@ class DcnmIntf:
                         )
                         continue
 
-                if str(have["deletable"]).lower() == "false":
+                if (
+                    str(have.get("deletable")).strip().lower()
+                    == "false"
+                ):
                     source = self.dcnm_intf_get_underlay_policy_source(have)
 
-                    if source is None:
+                    if source is not None:
+                        # Add this 'have to a deferred list. We will process this list once we have processed all the 'haves'
+                        defer_list.append(have)
+                        self.changed_dict[0]["deferred"].append(
+                            {
+                                "Name": name,
+                                "Deletable": have["deletable"],
+                                "Underlay Policies": have["underlayPolicies"],
+                                "Source": source,
+                            }
+                        )
+                        continue
+
+                    if self.module.params["state"] != "deleted":
                         self.dcnm_intf_skip_non_resolvable_deferred(have)
                         continue
 
-                    # Add this 'have to a deferred list. We will process this list once we have processed all the 'haves'
-                    defer_list.append(have)
-                    self.changed_dict[0]["deferred"].append(
-                        {
-                            "Name": name,
-                            "Deletable": have["deletable"],
-                            "Underlay Policies": have["underlayPolicies"],
-                            "Source": source,
-                        }
-                    )
-                    continue
+                    if not self.dcnm_intf_capability_enabled(
+                        have, "editAllowed"
+                    ):
+                        self.dcnm_intf_skip_physical_default_not_allowed(have)
+                        continue
 
                 uelem = self.dcnm_intf_get_default_eth_payload(
                     name, sno, fabric
@@ -6089,12 +6134,21 @@ class DcnmIntf:
                                 )
                             ):
 
-                                if (
-                                    str(match_have["deletable"]).lower()
-                                    == "false"
-                                ):
+                                deletable = self.dcnm_intf_capability_enabled(
+                                    match_have, "deletable"
+                                )
+                                edit_allowed = self.dcnm_intf_capability_enabled(
+                                    match_have, "editAllowed"
+                                )
+                                if not deletable and not edit_allowed:
+                                    self.dcnm_intf_skip_physical_default_not_allowed(
+                                        match_have
+                                    )
                                     continue
 
+                                using_edit_allowed = (
+                                    not deletable and edit_allowed
+                                )
                                 uelem = self.dcnm_intf_get_default_eth_payload(
                                     intf["ifName"],
                                     intf["serialNumber"],
@@ -6124,9 +6178,26 @@ class DcnmIntf:
                                         match_have
                                     )
                                     if rc is True:
-                                        if self.dcnm_intf_should_defer_deleted_member_default(
-                                            match_have, iface
+                                        defer_member_default = (
+                                            self.dcnm_intf_should_defer_deleted_member_default(
+                                                match_have, iface
+                                            )
+                                        )
+                                        source = (
+                                            self.dcnm_intf_get_underlay_policy_source(
+                                                match_have
+                                            )
+                                        )
+                                        if (
+                                            using_edit_allowed
+                                            and source is not None
+                                            and not defer_member_default
                                         ):
+                                            self.dcnm_intf_skip_edit_allowed_underlay_dependency(
+                                                match_have
+                                            )
+                                            continue
+                                        if defer_member_default:
                                             self.dcnm_intf_defer_deleted_member_default(
                                                 intf["ifName"],
                                                 intf["serialNumber"],
--- a/plugins/modules/dcnm_interface.py
+++ b/plugins/modules/dcnm_interface.py
@@ -5679,10 +5679,32 @@ class DcnmIntf:
                         )
                         continue
 
-                if (
-                    str(have.get("deletable")).strip().lower()
-                    == "false"
-                ):
+                is_deleted = self.module.params["state"] == "deleted"
+                deletable = self.dcnm_intf_capability_enabled(
+                    have, "deletable"
+                )
+                edit_allowed = self.dcnm_intf_capability_enabled(
+                    have, "editAllowed"
+                )
+
+                # A bulk deleted request must make the same fail-closed
+                # capability decision as a named deleted request. Missing or
+                # unrecognized controller metadata is not authorization to
+                # reset a physical interface.
+                if is_deleted and not deletable and not edit_allowed:
+                    self.dcnm_intf_skip_physical_default_not_allowed(have)
+                    continue
+
+                raw_deletable_is_false = (
+                    str(have.get("deletable")).strip().lower() == "false"
+                )
+                needs_dependency_handling = (
+                    not deletable
+                    if is_deleted
+                    else raw_deletable_is_false
+                )
+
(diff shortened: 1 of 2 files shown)


plugins/modules/dcnm_interface.py
```diff
@@ -5467,6 +5467,10 @@ def dcnm_intf_get_underlay_policy_source(self, intf):
 
         return None
 
+    def dcnm_intf_capability_enabled(self, intf, capability):
```
mikewiebe, line 5781: ### High: Require an affirmative reset capability in every bulk deleted path

**Issue**

The named-interface path safely requires either `deletable` or `editAllowed`
to normalize to true before it resets a physical Ethernet interface. The
delete-all and switch-only paths do not apply the same rule. They enter the
capability checks only when the raw `deletable` value is exactly false. If the
field is missing, `None`, blank, `0`, or an unrecognized controller value, the
code falls through and queues the role-default replacement without checking
whether `editAllowed` is affirmatively true.

For example, assume NDFC returns this record for a production server port:

```json
{
  "ifName": "Ethernet1/48",
  "isPhysical": true,
  "deletable": null,
  "editAllowed": false,
  "interfaces": [
    {
      "nvPairs": {
        "MODE": "trunk",
        "ALLOWED_VLANS": "100,200",
        "DESC": "Production server"
      }
    }
  ]
}
```

Neither capability affirmatively authorizes a reset: `deletable` is unknown
and `editAllowed` is false. A named delete handles this safely:

```yaml
- name: Reset one interface
  cisco.dcnm.dcnm_interface:
    fabric: Fabric-1
    state: deleted
    config:
      - name: Ethernet1/48
        type: eth
        switch:
          - 10.1.1.10
```

It skips the port with `changed: false` and the reason that neither
`deletable` nor `editAllowed` is true. Now run the equivalent bulk operation:

```yaml
- name: Reset every physical Ethernet interface
  cisco.dcnm.dcnm_interface:
    fabric: Fabric-1
    state: deleted
    override_intf_types:
      - eth
    config: []
    deploy: true
```

The bulk condition effectively evaluates:

```python
str(None).strip().lower() == "false"
# "none" == "false" -> False
```

Because it is false, the code bypasses the capability block and can return
`changed: true`, queue `Ethernet1/48` in the replacement payload, and report
no skipped interface. If deployed and accepted, that replacement can remove
the production trunk intent and apply the role-default policy; if the
controller rejects the unauthorized write, the broad operation can fail or
partially deploy.

| Request | Same capability record | Result |
| --- | --- | --- |
| Named delete | `deletable: null`, `editAllowed: false` | Safely skipped |
| Bulk delete | `deletable: null`, `editAllowed: false` | Reset incorrectly queued |

**Evidence**

- [`dcnm_intf_capability_enabled()` lines
  5470-5472](https://github.com/CiscoDevNet/ansible-dcnm/pull/710/files#diff-08575ad0f483bf55d1059fb810be34592862f247ea7d838fb7acfaa782b0f21fR5470)
  is fail-closed: only an explicit normalized `"true"` enables a capability.
- The named path uses that helper for both fields and skips when neither is
  true at [lines
  6137-6147](https://github.com/CiscoDevNet/ansible-dcnm/blob/c9edade5861dfbdd7b4cb5f73e5263ab21331e05/plugins/modules/dcnm_interface.py#L6137-L6147).
- The bulk path wraps all capability handling in a raw
  `str(have.get("deletable")).strip().lower() == "false"` condition at
  [lines
  5682-5710](https://github.com/CiscoDevNet/ansible-dcnm/blob/c9edade5861dfbdd7b4cb5f73e5263ab21331e05/plugins/modules/dcnm_interface.py#L5682-L5710).
  Missing or unknown values bypass the condition and proceed to default
  payload generation.
- A full current-head module-harness probe set every selected physical
  interface to `deletable=None` and `editAllowed=False`, then ran delete-all
  with `override_intf_types: [eth]`. It returned `changed: true`, an empty
  skipped list, and replacements for `Ethernet1/1`, `Ethernet1/2`, and
  `Ethernet3/2`.
- The equivalent named test with both capabilities false or with
  `editAllowed` missing is safely skipped. The added delete-all and
  switch-only tests exercise only `deletable=False` plus
  `editAllowed=True`.

**Existing PR overlap**

**Impact**

A broad deleted operation can enqueue physical-interface policy writes when
the controller did not affirm that deletion or editing is allowed. Missing
metadata can result from a version contract change, partial response, or
fixture drift and must not become mutation authorization. The blast radius is
larger than the named path because delete-all can target every physical port
on every selected switch.

**Suggested fix**

Separate the controller metadata from the requested state:

1. Normalize both capabilities through `dcnm_intf_capability_enabled()`.
2. For physical reset under `state: deleted`, continue only when
   `deletable or editAllowed` is true; otherwise add the explicit skipped
   record.
3. Preserve the existing underlay-member dependency and deferral ordering once
   reset permission has been established.
4. Keep `state: overridden` semantics separately scoped so a change intended
   for deleted state does not silently broaden it.

Add delete-all and switch-only tests for missing, `None`, blank, unrecognized,
and both-false values, plus boolean/string true variants. Assert no
replacement or deployment is queued without affirmative capability, and test
named and bulk requests from the same fixture so their decisions stay
consistent.

---

tests/unit/modules/dcnm/test_dcnm_intf.py
```diff
@@ -3978,6 +4049,172 @@ def test_dcnm_intf_eth_deleted_existing(self):
...
+                fabric="test_fabric",
+                config=self.playbook_config,
+            )
+        )
+        result = self.execute_module(changed=True, failed=False)
+
+        self.assertEqual(len(result["diff"][0]["replaced"]), 5)
+        self.assertEqual(len(result["diff"][0]["skipped"]), 0)
+
+    def test_dcnm_intf_eth_deleted_existing_nd42_check_mode(self):
```
mikewiebe, line 4671: ### Medium: Report a planned physical reset as changed in check mode

**Issue**

The PR adds a check-mode test for the newly enabled ND 4.2 physical-reset path,
but it codifies `changed: false` while also asserting five planned
replacements. Check mode is supposed to suppress controller writes while
reporting whether a real run would make a change.

For example, the new test's result has this contradictory shape:

```yaml
changed: false
diff:
  - replaced:
      - Ethernet1/1 role-default payload
      - Ethernet1/2 role-default payload
      # three more replacements
response: []
```

Automation using check mode cannot distinguish “already default” from “five
resets would be made.”

**Evidence**

- [`test_dcnm_intf_eth_deleted_existing_nd42_check_mode()` lines
  4093-4109](https://github.com/CiscoDevNet/ansible-dcnm/pull/710/files#diff-442ae9a22bd84bc32231094bcfa014d65e17fea8b7adaf9fe3a4a87a04786c74R4093)
  explicitly expects `changed=False` while asserting five replacements and no
  response.
- The module first sets `changed=True` when a replacement diff exists at
  [lines
  7097-7113](https://github.com/CiscoDevNet/ansible-dcnm/blob/c9edade5861dfbdd7b4cb5f73e5263ab21331e05/plugins/modules/dcnm_interface.py#L7097-L7113),
  then overwrites it with false solely because check mode is active at
  [lines
  7115-7117](https://github.com/CiscoDevNet/ansible-dcnm/blob/c9edade5861dfbdd7b4cb5f73e5263ab21331e05/plugins/modules/dcnm_interface.py#L7115-L7117).
- The focused current-head test run passes, confirming that the contradictory
  expectation is now checked in.

**Existing PR overlap**

**Impact**

The PR's new reset workflow appears harmless during a dry run even when the
real task will replace physical-interface policies. CI will also protect the
wrong expectation, making it harder to correct later. This affects change
planning, approval gates, and idempotency diagnostics.

**Suggested fix**

Do not overwrite an already-computed `changed=True` result in check mode.
Exit before `dcnm_intf_send_message_to_dcnm()` while retaining the planned
changed result and diff. Change the new test to expect `changed=True`,
`response: []`, and the exact replacement payloads. Cover named, delete-all,
and switch-only ND 4.2 reset cases plus an already-default negative case that
correctly returns `changed=False`.

dacasti2mx:
Problem reproduced, validated this fix against NDFC 12.5.0.475 using an unused physical Ethernet interface with:

  ```text
  deletable=false
  editAllowed=true
```
  With the pre-fix code, state: deleted returned changed=false and left the interface configuration unchanged, reproducing
  issue #709.

  With this PR:
```bash
  - Check mode correctly reported changed=true.
  - Removing the interface through the ansible-dc-vxlan diff-run workflow reset it to the default policy.
  - NDFC accepted the PUT /rest/interface request with HTTP 200.
  - An independent query confirmed that the custom description was cleared.
  - A second run returned changed=false, confirming idempotency.
```
  This live test confirms the named physical-interface delete path. Bulk and switch-only deletes are covered too.

dacasti2mx (approved):
LGTM

PR #661, opened 2026-03-30
Add FEC on dcnm_interface for ND4.1

ccoueffe:
#660 

Use existing plugin: version_compare to compare and push only if ND >= 4.1.1

```
---
- name: test inventory
  hosts: nac-ndfc1
  gather_facts: false

  tasks:

  - name: Configure 25G with fc-fec on E1/12
    cisco.dcnm.dcnm_interface:
      fabric: "nac-ndfc1"
      state: merged
      config:
        - name: eth1/12
          type: eth
          switch:
            - "10.229.42.121"
          deploy: false
          profile:
            mode: trunk
            admin_state: true
            speed: 25Gb
            fec: rs-fec
            description: "ETH 1/12 25G fc-fec"
```

```
TASK [Configure 25G with fc-fec on E1/12] *************************************************************************************************
task path: /Users/ccoueffe/Documents/Devnet/NDFC-Ansible-As-Code/ndfc-ansbile-as-code/nac-ndfc/playbooks/play-interface_fec.yaml:8
Monday 30 March 2026  14:45:06 +0200 (0:00:00.053)       0:00:00.053 ********** 
Monday 30 March 2026  14:45:06 +0200 (0:00:00.053)       0:00:00.053 ********** 
changed: [nac-ndfc1] => {
    "changed": true,
    "diff": [
        {
            "debugs": [],
            "deferred": [],
            "delete_deploy": [],
            "deleted": [],
            "deploy": [],
            "merged": [
                {
                    "interfaceType": "INTERFACE_ETHERNET",
                    "interfaces": [
                        {
                            "ifName": "Ethernet1/12",
                            "nvPairs": {
                                "FEC": "rs-fec"
                            }
                        }
                    ]
                }
            ],
            "overridden": [],
            "query": [],
            "replaced": [],
            "skipped": []
        }
    ],
    "invocation": {
        "module_args": {
            "check_deploy": false,
            "config": [
                {
                    "deploy": false,
                    "name": "eth1/12",
                    "profile": {
                        "admin_state": true,
                        "description": "ETH 1/12 25G fc-fec",
                        "fec": "rs-fec",
                        "mode": "trunk",
                        "speed": "25Gb"
                    },
                    "switch": [
                        "10.229.42.121"
                    ],
                    "type": "eth"
                }
            ],
            "deploy": true,
            "fabric": "nac-ndfc1",
            "override_intf_types": [],
            "state": "merged"
        }
    },
    "response": [
        {
            "DATA": {},
[...]

PR #712, opened 2026-07-29
Fix post-merge FEC and XConnect review findings

mikewiebe:
## Summary

Addresses outstanding review feedback after PR #661 and PR #612 were merged into `develop`.

## PR #661: dcnm_interface FEC

- Reuses the existing controller-version response instead of issuing a second version request.
- Preserves the exact normalized NDFC version needed for the NDFC 12.4.1 FEC boundary.
- Reports malformed and failed version responses with actionable response context.
- Adds focused coverage for supported, older, unknown, malformed, and failed version responses.
- Adds regression coverage for exact FEC payloads, role-default reset behavior, FEC-only drift, named deleted, overridden, check mode, and idempotence.

## PR #612: dcnm_network XConnect

- Carries the exact NDFC version from the action plugin to the module through `_fabric_details`.
- Removes the duplicate module-level controller-version request.
- Preserves controller-returned `xconnect` intent when the option is omitted from an unrelated merged operation.
- Retains XConnect in normalized and formatted output whenever the controller returns the field.
- Keeps an omitted `xconnect` value unset during input validation so older and unknown NDFC versions are not rejected as though `xconnect: false` had been explicitly configured.
- On NDFC 12.4.1 and later, serializes an omitted `xconnect` value as JSON `false` at the network payload boundary instead of sending `null`. This prevents the NaC bulk-create failure `Invalid values provided for these parameters: [xconnect: null]`.
- Documents the standalone-fabric, Layer-2-only, ND 4.1/NDFC 12.4.1 minimum, and dot1q attachment contract.
- Adds a complete standalone XConnect example and focused validation, payload, preservation, output, and idempotence tests.
- Adds regression coverage for omitted `xconnect` on unknown, NDFC 11.1, and NDFC 12.2.1.321 versions, plus exact boolean serialization for the NDFC 12.4.1+ payload.

## Validation

- `263 passed, 61 subtests passed`
- `ansible-test sanity --test validate-modules`: passed
- Python 3.12 compile sanity: passed
- Pep8 sanity: passed
- `git diff --check`: passed

The `xconnect: null` failure was observed through NaC against a live NDFC deployment. No additional end-to-end live ND run was performed as part of this update.

### DIFF
--- a/plugins/action/dcnm_network.py
+++ b/plugins/action/dcnm_network.py
@@ -116,6 +116,7 @@ class ActionModule(ActionBase):
         self.logger.info("NDFC Network Action Plugin initialized")
         # Default NDFC version
         self.ndfc_version = 12.2
+        self.ndfc_full_version = None
 
     def run(self, tmp=None, task_vars=None):
         """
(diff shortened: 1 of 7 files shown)


juburnet (approved):
Approved.  This fixed the issues we saw in our pipeline regarding dcnm_network xconnect.

PR #716, opened 2026-08-19
Handle secondary gateway payloads by NDFC version

### LINKED ISSUE #715: dcnm_network: secondaryGWs breaks network creation on NDFC 12.2.2.238
### Community Note

* Please vote on this issue by adding a 👍 reaction to the original issue to help the community and maintainers prioritize it.
* Please do not leave "+1" comments that do not add relevant information.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

### Summary

`cisco.dcnm` 3.12.1 adds the aggregated `secondaryGWs` field to `networkTemplateConfig` when creating or processing networks.

NDFC 12.2.2.238, running on ND 3.2(1e), rejects the network creation request when this field is present, even when no secondary gateways are configured.

The same network can be created successfully when `secondaryGWs` is omitted and the legacy `secondaryGW1` through `secondaryGW4` fields are preserved.

### Ansible Version and Collection Version

Affected environment:

```text
ansible-core: 2.16.3
Python: 3.10.13
cisco.dcnm: 3.12.1
```

Working control:

```text
cisco.dcnm: 3.11.1
```

The issue was originally observed through `cisco.nac_dc_vxlan`, but the payload comparison and controlled field-removal test locate the incompatible behavior in `dcnm_network`.

### NDFC Version

Affected:

```text
ND: 3.2(1e)
NDFC: 12.2.2.238
```

Positive control:

```text
ND: 3.2(2m)
NDFC: 12.2.3.70
```

The exact first supported NDFC build still needs to be confirmed.

### Affected Module

```text
cisco.dcnm.dcnm_network
```

The behavior is present in `plugins/modules/dcnm_network.py` from the `3.12.1` tag.

`secondaryGWs` is added in four processing paths:

```text
update_create_params()
normalize_have_network()
push_to_remote()
dcnm_update_network_information()
```

The relevant locations in the 3.12.1 tag are approximately:

```text
2513-2514
2566-2567
4812-4813
5571-5572
```

### Potential Fabric Scope

The live failure was confirmed on a multicluster parent.

However, the current condition adds `secondaryGWs` to every fabric type except:

```text
multisite_child
multicluster_child
```

Based on code inspection, the affected path is also reachable for:

* standalone VXLAN EVPN fabrics;
* standalone eBGP VXLAN fabrics;
* multisite parent fabrics; and
* multicluster parent fabrics.

These additional fabric types have not yet been reproduced against NDFC 12.2.2.238, so their runtime impact should be treated as potentially affected rather than confirmed.

### Minimal Reproduction

Create an L3 network on a multicluster parent without configuring any secondary gateway:

```yaml
config:
  - net_name: test_network
    vrf_name: test_vrf
    net_id: 4200002
    vlan_id: 2000
    vlan_name: test_network
    gw_ip_subnet: 169.254.0.129/25
    arp_suppress: true
    deploy: false
```

The affected execution uses:

```yaml
state: replaced
fabric: MCFG_PARENT
```

### Payload Comparison

With `cisco.dcnm` 3.11.1, the generated `networkTemplateConfig` contains only the legacy fields:

```json
{
  "secondaryGW1": "",
  "secondaryGW2": "",
  "secondaryGW3": "",
  "secondaryGW4": ""
}
```

With `cisco.dcnm` 3.12.1, the same logical network contains one additional field:

```json
{
  "secondaryGW1": "",
  "secondaryGW2": "",
  "secondaryGW3": "",
  "secondaryGW4": "",
  "secondaryGWs": "{\"secondaryGWs\":[]}"
}
```

After normalizing and comparing both captured requests, `secondaryGWs` was the only semantic difference inside `networkTemplateConfig`.

The complete request grew from 918 to 967 bytes. The 49-byte increase matches the serialized and escaped `secondaryGWs` field exactly.

### Actual Behavior

NDFC 12.2.2.238 rejects the network creation request:

```text
HTTP 400 Bad Request

Invalid template config parameters:
invalid character 's' after object key:value pair
```

The failure occurs on:

```text
POST /onemanage/appcenter/cisco/ndfc/api/v1/onemanage/top-down/fabrics/<fabric>/networks
```

This happens before the network attachment request.

### Controlled Confirmation

A controlled test was performed with `cisco.dcnm` 3.12.1 on the affected controller:

1. The original 3.12.1 code generated `secondaryGWs`.
2. Network creation failed with HTTP 400.
3. Only the four code paths that add `secondaryGWs` were disabled.
4. The same network was submitted again.
5. Network creation succeeded.
6. Additional networks were also created successfully.

No secondary gateways were configured or required in this scenario.

### Expected Behavior

`dcnm_network` should generate a network payload compatible with the connected NDFC version.

For controllers that do not support the aggregated `secondaryGWs` field, the module should:

* omit `secondaryGWs`;
* preserve the legacy `secondaryGW1` through `secondaryGW4` fields; and
* continue allowing networks without secondary gateways to be created.

For supported controllers, `secondaryGWs` should continue to be generated.

If a user requests more secondary gateways than an older controller can represent, the module should fail clearly instead of silently dropping the unsupported intent.

### Suggested Fix Boundary

Gate `secondaryGWs` on controller support in all four processing paths. Preserve the legacy fields on older controllers and fail clearly when the requested configuration cannot be represented safely.

### Suggested Tests

* Verify that NDFC 12.2.2.238 receives only the legacy secondary gateway fields.
* Verify that supported controllers continue receiving `secondaryGWs`.
* Verify create, update and idempotency with and without secondary gateways.
* Cover standalone and parent fabric paths.

### References

Related historical support:

* #196
* #208

Those references added secondary gateway functionality but do not cover this controller-version compatibility behavior.

dacasti2mx:
### Summary

NDFC 12.2.2.238 does not support the aggregate `secondaryGWs` network-template field, while the
known-compatible 12.2.3.70 release does. `dcnm_network` previously generated the aggregate field for
standalone and parent fabrics without considering the controller version.

This change makes secondary-gateway payload generation version-aware while preserving the four
existing `secondaryGW1`-`secondaryGW4` fields:

- versions before the minimum known-compatible 12.2.3.70 omit `secondaryGWs`;
- 12.2.3.70 and newer versions generate `secondaryGWs`;
- unknown versions fail before a write when gateways are configured;
- child fabrics retain their existing behavior;
- controller state that cannot be represented safely by the four positional fields fails closed;

The module uses the normalized full version already supplied by the existing action/helper flow;
no new raw-version field or version lookup was added.

Fixes #715.

### Validation

- Focused secondary-gateway tests: 68 passed, 173 subtests passed.
- Complete `dcnm_network` unit suite: 183 passed, 184 subtests passed.
- Complete unit suite: 3076 passed, 234 subtests passed; the single known suite-order logging test
  passed when run alone.
- `py_compile`, materialized pep8, and `git diff --check`: passed.
- `validate-modules`: no new findings; only the repository's existing license/RETURN findings.
- Live create on NDFC 12.2.3.70: wire capture confirmed `secondaryGWs` was sent.
- Isolated forced-legacy control on the same lab: wire capture confirmed `secondaryGWs` was omitted
  while `secondaryGW1`-`secondaryGW4` were preserved.
- Create, deploy, reversible update, idempotency, ordered cleanup, NDFC verification, and NX-OS
  verification passed; the original lab baseline was restored.

### Scope and limitations

- 12.2.3.70 is the minimum version confirmed compatible by this investigation; the conservative
  interval below it is treated as legacy.

### Files changed

- `plugins/modules/dcnm_network.py`
- `tests/unit/module_utils/common/test_dcnm_version_supported.py`
- `tests/unit/modules/dcnm/test_dcnm_network.py`

### DIFF
--- a/plugins/modules/dcnm_network.py
+++ b/plugins/modules/dcnm_network.py
@@ -264,21 +264,26 @@ options:
       secondary_ip_gw1:
         description:
         - IP address with subnet for secondary gateway 1
+        - NDFC versions from the minimum known-compatible 12.2.3.70 use an aggregate list and require gateways in slot order without gaps
+        - Older versions keep the four positional fields; an unknown version with configured gateways fails before any write
         type: str
         required: false
       secondary_ip_gw2:
         description:
         - IP address with subnet for secondary gateway 2
+        - See secondary_ip_gw1 for version compatibility and slot ordering
         type: str
         required: false
       secondary_ip_gw3:
         description:
         - IP address with subnet for secondary gateway 3
+        - See secondary_ip_gw1 for version compatibility and slot ordering
         type: str
         required: false
       secondary_ip_gw4:
         description:
         - IP address with subnet for secondary gateway 4
+        - See secondary_ip_gw1 for version compatibility and slot ordering
         type: str
         required: false
       trm_enable:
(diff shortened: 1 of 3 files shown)


mikewiebe:
### Low: Avoid quadratic desired-network scanning

**Issue**

For `state: overridden`, each existing network scans the full desired-network list to determine whether it should be retained. This creates O(existing × desired) processing.

**Evidence**

[`plugins/modules/dcnm_network.py:2887`](https://github.com/CiscoDevNet/ansible-dcnm/pull/716/files#diff-81913290492fc2a4c62ca8ade79047040d6afc496e3dbc92761e14a1ed6f472cR2887) calls `any(...)` over `self.want_create` inside `normalize_have_network()`.

A benchmark on this PR head showed approximately quadratic growth:

- 800 × 800: 0.0302 seconds
- 1,600 × 1,600: 0.1069 seconds
- 3,200 × 3,200: 0.3993 seconds

**Existing PR overlap**

**Impact**

Large overridden operations may spend increasingly significant time performing repeated name lookups.

**Suggested fix**

Build a set or dictionary of desired network names once after `get_want()`, then use constant-time membership checks during normalization. Add a scale-oriented overridden-state unit test.

dacasti2mx:
> ### Low: Avoid quadratic desired-network scanning
> 
> **Issue**
> 
> For `state: overridden`, each existing network scans the full desired-network list to determine whether it should be retained. This creates O(existing × desired) processing.
> 
> **Evidence**
> 
> [`plugins/modules/dcnm_network.py:2887`](https://github.com/CiscoDevNet/ansible-dcnm/pull/716/files#diff-81913290492fc2a4c62ca8ade79047040d6afc496e3dbc92761e14a1ed6f472cR2887) calls `any(...)` over `self.want_create` inside `normalize_have_network()`.
> 
> A benchmark on this PR head showed approximately quadratic growth:
> 
>     * 800 × 800: 0.0302 seconds
> 
>     * 1,600 × 1,600: 0.1069 seconds
> 
>     * 3,200 × 3,200: 0.3993 seconds
> 
> 
> **Existing PR overlap**
> 
> No matching existing PR comment found.
> 
> **Impact**
> 
> Large overridden operations may spend increasingly significant time performing repeated name lookups.
> 
> **Suggested fix**
> 
> Build a set or dictionary of desired network names once after `get_want()`, then use constant-time membership checks during normalization. Add a scale-oriented overridden-state unit test.

Thanks @mikewiebe, working on that

PR #724, opened 2026-09-15
Management interface support

mthurstocisco:
Just as above


plugins/modules/dcnm_interface.py
```diff
@@ -970,6 +970,47 @@
...
+          or removed during 'deleted' and 'overridden' states.
+        - This profile requires NDFC/DCNM 12 or above, since it uses the 'int_mgmt' policy.
+        suboptions:
+          mode:
+            description:
+            - Interface mode
+            choices: ['mgmt']
+            type: str
+            required: true
+          cmds:
```
mikewiebe, line 990: ### Medium: Protect existing management `CONF` when `cmds` is omitted

**Issue**

A switch managed by NDFC must already be reachable through its management
address. Changing that active address through the same controller is inherently
disruptive and would require a coordinated switch and NDFC-inventory migration;
the normal safe use of this profile is therefore to retain the existing
management `CONF` while changing metadata such as description or CDP.

However, `cmds` is optional, and `state: replaced` turns an omitted value into
`CONF: ""`. Although replacement state ordinarily expects a complete desired
resource, the public management-profile contract neither requires `cmds` nor
warns that omission can clear the freeform configuration, and deployment
defaults to `true`. For `mgmt0`, that ambiguity can produce destructive intent
without the user explicitly requesting an address change or removal.

**Evidence**

- [`profile_mgmt.cmds` at `plugins/modules/dcnm_interface.py` line 990](https://github.com/CiscoDevNet/ansible-dcnm/pull/724/files#diff-08575ad0f483bf55d1059fb810be34592862f247ea7d838fb7acfaa782b0f21fR990)
  says the management address is carried as a freeform command but leaves the
  field optional with `default: []`.
- [`dcnm_intf_get_mgmt_payload()` at line 4416](https://github.com/CiscoDevNet/ansible-dcnm/pull/724/files#diff-08575ad0f483bf55d1059fb810be34592862f247ea7d838fb7acfaa782b0f21fR4416)
  serializes the normalized omitted value (`None`) as an empty `CONF`.
- The PR's [`test_dcnm_intf_mgmt_payload_no_cmds` at line 6050](https://github.com/CiscoDevNet/ansible-dcnm/pull/724/files#diff-442ae9a22bd84bc32231094bcfa014d65e17fea8b7adaf9fe3a4a87a04786c74R6050)
  explicitly asserts that omission produces `CONF == ""`.
- The integration-test preamble explicitly says that the tests [never change
  the address, resend the existing `mgmt_conf` with every request, and use
  `deploy: false`](https://github.com/CiscoDevNet/ansible-dcnm/pull/724/files#diff-abdcb8ea1893d4fdce5c0fd3f66ec6c4ab683df1bf651037a6c48cfa71e310a2R5-R9).
  Those safeguards are not enforced by the module's public contract.
- The common interface contract leaves [`deploy` enabled by default](https://github.com/CiscoDevNet/ansible-dcnm/blob/544689f3d875009a290594178af009b7dacade0d/plugins/modules/dcnm_interface.py#L102-L107).
- A full module-harness probe removed `cmds` from the supplied
  `mgmt_replaced_config`; the resulting create/update request contained
  `nvPairs.CONF == ""`. This confirms the outbound request only; the payload
  has not been deployed on a live controller.

For example, assume NDFC already reaches a switch through:

```text
interface mgmt0
  description Old description
  ip address 10.122.84.181/24
  no shutdown
```

An operator replacing only the metadata might provide:

```yaml
- name: Replace management-interface metadata
  cisco.dcnm.dcnm_interface:
    fabric: PROD-FABRIC
    state: replaced
    config:
      - name: mgmt0
        type: mgmt
        switch:
          - 10.122.84.181
        profile:
          mode: mgmt
          description: Managed by Ansible
          enable_cdp: true
```

The user has not requested an address change, but the module builds:

```yaml
policy: int_mgmt
interfaces:
  - ifName: Mgmt0
    nvPairs:
      INTF_NAME: Mgmt0
      DESC: Managed by Ansible
      CONF: ""
      ADMIN_STATE: "true"
      CDP_ENABLE: "true"
```

**Existing PR overlap**

No duplicate finding found. This is separate from the [`int_mgmt` policy
priority finding](https://github.com/CiscoDevNet/ansible-dcnm/pull/724#discussion_r4030053768),
which can occur even when the existing `CONF` is supplied correctly.

**Impact**

If NDFC interprets empty `CONF` as replacement intent and deploys it, an
ordinary metadata update can remove the command that provides controller
reachability. If NDFC preserves the existing configuration independently, that
device impact does not occur, but the module contract remains ambiguous. Since
the trigger uses full replacement semantics and the controller behavior has
not yet been reproduced, this is Medium pending a controlled live test; the
operational consequence would be severe if address removal is confirmed.

**Suggested fix**

Treat management `CONF` as protected reachability configuration. Distinguish
an omitted `cmds` field from an explicitly empty list. When it is omitted,
either preserve the current `CONF` even for an explicitly named management
interface under replacement-style processing, or fail before any mutation
with an actionable message requiring the existing complete `CONF`.

Do not add a dedicated management-IP property merely to solve this omission
case. If an explicit empty or changed management-address command is supported,
make it a deliberate opt-in operation with a clear connectivity warning and a
documented coordinated NDFC-inventory workflow; otherwise reject it. Add tests
for omitted versus explicit-empty `cmds` in `replaced` and explicit
`overridden` input, validation before mutation, check mode, exact outbound
payloads, and idempotent readback.

Finally, run a controlled live ND test on a recoverable switch: capture the
current `int_mgmt` intent and running `mgmt0` configuration, replace only
description/CDP while omitting `cmds`, deploy, verify whether NDFC preserves or
removes the address and whether controller reachability survives, then restore
the original configuration.

---

plugins/modules/dcnm_interface.py
```diff
@@ -5056,9 +5207,73 @@ def dcnm_intf_compare_want_and_have(self, state):
                             # the serial number should be unique across all fabrics
                             if_keys.remove("fabricName")
                             changed_dict[k][0].pop("fabricName")
+
+                            # 'int_mgmt' policies on bootstrap provisioned switches carry nvPairs
```
mikewiebe, line 5211: **@mthurstocisco This is one possible way to handle the priority change but would need to test/evaluate on a live ND instance and verify behavior on ND4.1 -> ND4.3**

##### Medium: Preserve the `int_mgmt` policy priority across interface updates

**Issue**

The management-interface comparison copies controller-owned nvPairs such as
`PRIORITY` and `POLICY_ID` into the outgoing interface request, and the new
unit test treats that as proof that the metadata survives. The PR originator
reports, however, that on ND 3.2 the GUI-style interface update recreates the
underlying `int_mgmt` policy with priority `500` regardless of the value sent.
`dcnm_interface` does not read the recreated policy back or restore its
top-level priority through the policy endpoint, so an otherwise successful
description, CDP, or `CONF` update can silently change policy metadata.

**Evidence**

- [The management preservation block at
  `plugins/modules/dcnm_interface.py` line 5211](https://github.com/CiscoDevNet/ansible-dcnm/pull/724/files#diff-08575ad0f483bf55d1059fb810be34592862f247ea7d838fb7acfaa782b0f21fR5211)
  copies every nvPair absent from the desired payload, including `PRIORITY`
  and `POLICY_ID`, and says this prevents the controller from dropping them.
- [The management fixture at
  `tests/unit/modules/dcnm/fixtures/dcnm_intf_mgmt_payloads.json` line
  28](https://github.com/CiscoDevNet/ansible-dcnm/pull/724/files#diff-2aac4295d89c9f89155a67117b385ae545c95342b4f1cd6b5f12e0bfe009dfb1R28)
  models a bootstrap policy with `POLICY_ID: POLICY-2020` and `PRIORITY:
  "900"`.
- [`test_dcnm_intf_mgmt_preserve_unmanaged_nvpairs` at line
  6233](https://github.com/CiscoDevNet/ansible-dcnm/pull/724/files#diff-442ae9a22bd84bc32231094bcfa014d65e17fea8b7adaf9fe3a4a87a04786c74R6233)
  checks only that the mocked outbound request contains priority `900`; its
  mocked controller always returns success and there is no post-update policy
  readback.
- Existing-interface changes are sent through the interface APIs
  (`/interface/modify` or `/interface`), while new intent uses
  [`/globalInterface`](https://github.com/CiscoDevNet/ansible-dcnm/blob/544689f3d875009a290594178af009b7dacade0d/plugins/modules/dcnm_interface.py#L7294-L7355).
  This module never calls the separate `/control/policies/...` update API.
- The collection already implements policy modification as a full-policy
  `PUT` to
  [`/control/policies/{policyId}/bulk`](https://github.com/CiscoDevNet/ansible-dcnm/blob/544689f3d875009a290594178af009b7dacade0d/plugins/modules/dcnm_policy.py#L1142-L1150).
- The reported ND 3.2 behavior has not yet been independently reproduced in
  this review. A live-controller test is required before choosing the final
  update sequence.

For example, a description-only update can appear to preserve priority in the
request while changing the stored policy:

```text
Before:            POLICY-2020, priority 900, description "Original"
Interface request: PRIORITY=900,          description "Managed by Ansible"
Readback on ND:    POLICY-3487, priority 500, description "Managed by Ansible"
```

The management IP can remain intact in this example because the existing
`CONF` is supplied; the defect is the independent policy recreation and
priority drift.

**Existing PR overlap**

No matching existing PR comment found. This is related to, but does not
duplicate, the [active management-`CONF`
finding](https://github.com/CiscoDevNet/ansible-dcnm/pull/724#discussion_r4029570043):
that finding concerns a potentially empty management configuration, whereas
this issue can occur even when the complete existing `CONF` is sent.

**Impact**

A user can request an ordinary management-interface change and receive a
successful result while the policy moves from its existing priority to `500`.
That changes its ordering relative to policies with intervening priorities and
can affect precedence if their configuration overlaps. Even when no policy
overlaps, it creates unmanaged metadata drift that `dcnm_interface` cannot
restore. A subsequent run is likely to read `500`, copy it forward, and appear
idempotent, masking the original change. The concrete operational impact and
affected ND releases remain to be established on a live controller.

**Suggested fix**

First use a recoverable live ND test switch to determine whether an existing
`int_mgmt` policy can be updated directly through the policy API without being
recreated. Capture the full policy, its ID and top-level priority, modify only
the requested management nvPairs, update it through
`/control/policies/{policyId}/bulk`, deploy, and verify the ID, priority,
running configuration, reachability, and second-run idempotence. If that path
works, use it for existing management interfaces instead of the GUI-style
interface update; it avoids policy recreation entirely.

If the interface endpoint is required, preserve the original priority in a
separate per-interface snapshot, perform the interface update, re-query the
interface to discover its current policy ID, fetch the complete new policy,
restore only its top-level priority through the policy bulk-update endpoint,
verify the readback, and only then deploy. Do not attempt to restore the stale
old policy ID, and do not construct a partial policy payload that could discard
new controller fields. Include bounded retry handling for eventual consistency
and fail before deployment with an actionable error if priority restoration
cannot be confirmed.

Add tests covering direct-policy update or the recreate-and-repair fallback,
an original non-default priority, a changed policy ID, an already-`500`
policy, multiple management interfaces, policy-readback failure, restore
failure, check mode, and final idempotence. The decisive test must run against
live ND 3.2 and at least one other supported release because mocked request
inspection cannot verify controller-side preservation.

---

PR #725, opened 2026-09-17
dcnm_interface: table-driven nvPair support (generic interface engine)

dacasti2mx:
> **Draft, opened early on purpose.** The design is settled and validated on hardware, but
> the scope is large and I would rather take direction now than after another round. Comments
> on the approach are more useful to me than comments on the wording.

## The problem

`dcnm_interface` supports a fixed set of interface fields, each wired by hand: an entry in the
argument spec, a branch in the payload builder, a line in the compare path, and a place in the
`have` reconciliation. Adding one NDFC `nvPair` means touching four or five places and knowing
which of them apply to which of the eighteen interface templates.

The result is that most template fields are simply unsupported, and adding them does not get
cheaper with practice. `ACL_FILTER` and `GUARD_MODE` have existed in the NDFC templates for
releases; the module cannot express either.

## What this does

Introduces a **thin engine** driven by a **static binding table**. A binding declares what a
field is — parent template, `nvPair`, profile key, type, applicable interface type and mode,
minimum NDFC version, constraints — and the engine does the rest at five hook points in the
existing module. Adding a field becomes a table entry plus a test, not a code path.

```
tools/gie_generate_binding_table.py   reviewed registry YAML  ->  generated table
plugins/module_utils/gie_binding_table.py   672 lines, GENERATED, carries a provenance hash
plugins/module_utils/gie_engine.py          488 lines, the whole engine
```

Runtime never reads YAML and never queries NDFC template metadata to decide what a field is.
The table is generated offline from an approved registry and committed, so what ships is
exactly what was reviewed.

**Currently registered: 58 bindings across 9 parents, 14 distinct nvPairs.**

| Parent | Bindings |
|---|---|
| `int_trunk_host` | 9 |
| `int_access_host` | 8 |
| `int_port_channel_trunk_host` / `int_vpc_trunk_host` | 7 each |
| `int_port_channel_access_host` / `int_port_channel_dot1q_tunnel_host` / `int_routed_host` / `int_vpc_access_host` | 6 each |
| `int_fabric_loopback_11_1` | 3 |

55 are `passthrough` (the parent carries the nvPair directly); 3 are `child_pti` (the parent
propagates to a generated child policy), all three OSPF message-digest on fabric loopbacks.

## Design decisions worth arguing about

**Fail closed, everywhere.** An unknown binding type, an unregistered key on a parent, a value
of the wrong native type, a version that cannot be parsed — each aborts before anything is sent,
and the message says explicitly that no template metadata was queried and no change was made.

**Local validation runs before any GET.** The whole config is validated in one pass, so a
malformed field on interface #7 fails before interface #1 triggers a capability lookup. Tested.

**Messages never echo the rejected value.** A binding can carry a secret — the OSPF
authentication key is one — and module errors land in logs and CI artifacts. Errors name the
field, the expected type and the type received (`given a null value`), never the value, and
never its length.

**The empty string clears a plain-string binding.** NDFC returns `""` for an unset string and
accepts `""` to clear one; without this a field is settable and never clearable. Scoped to
bindings with no `valid_values`, so an empty enum still fails closed — for an enum, "off" is a
named choice, never the empty string.

**`str` subclasses are accepted.** Ansible hands a module `AnsibleUnicode`, not `str`. An exact
`type(v) is str` check rejects every string a real playbook can supply while unit tests passing
literals keep passing — that defect is in here as a fixed commit, and the check is now explicit
about `bool`/`int` not being interchangeable.

## Validation

Unit: **4021 pass** across the collection, including 9 new engine-specific suites — the thin
engine itself, passthrough wire form, boolean wire form, clearing a string binding, the
parent/child OSPF contract, "every registered parent is actually wired", "the legacy path and
the engine never write the same nvPair", routed bindings, and dot1q/QoS parity.

One test fails in the full run: `test_log_00011`. It passes in isolation and fails when run
alongside the rest of `test_log.py` — order dependence in the logging tests, present on
`develop` and untouched by this branch. Flagging it so it is not mistaken for fallout here.

Hardware: NDFC 12.6.0.267 / ND 4.3.1 / NX-OS 10.5(5) on a VXLAN EVPN fabric, exercising all
registered bindings on ethernet, port-channel, vPC and routed host interfaces plus the fabric
loopback. Set, change, omit, and clear were each measured on the device CLI rather than
inferred from a module `changed` flag — including a full withdrawal matrix (value present,
inert value, field omitted, whole interface removed from the model) and an end-to-end run
driven by a NaC data model rather than a direct playbook.

## What this deliberately does NOT do

- **No new interface types.** Only fields on parents the module already handles.
- **No composite fields.** `HSRP_SECONDARY_VIPS` and similar remain out; they need a shape the
  table does not express yet.
- **`query` state is not exercised** by the new tests. It is unchanged, but say so out loud.
- **The legacy hardcoded paths are not removed.** The engine runs alongside them and a test
  asserts the two never write the same nvPair. Migrating them is a follow-up, not this PR.

## Suggested review order

1. `plugins/module_utils/gie_engine.py` — the whole contract is here, 488 lines.
2. `tools/gie_generate_binding_table.py` and a handful of entries in the generated table.
3. `tests/unit/modules/dcnm/test_gie_thin_engine.py` — the engine's behaviour as specified.
4. The five hook points in `plugins/modules/dcnm_interface.py`, which is where I would most
   like a second opinion on placement.

Commits are ordered and each is self-contained; reviewing by commit is likely easier than by
file.

## Open questions I would like opinions on

1. **Hook placement.** The five hooks sit where they cause the least disruption today. Is that
   the right shape, or should the module call the engine from fewer, higher-level points?
2. **Table generation.** The table is generated and committed with a provenance hash. Is
   committing generated code acceptable here, or would you prefer the registry itself ship and
   be parsed at import?
3. **Scope of this PR.** 31 commits is a lot. I can split the engine from the binding
   registrations if that makes review tractable — say the word before you start reading.


mikewiebe:
### Advisory: Generic binding scale and carry-forward behavior

The PR appears to preserve the module’s main controller-facing performance improvements which is good:

- Interface data is still bulk-fetched and cached per switch.
- Updates still use the bulk API.
- The static binding registry does not introduce a REST request per property.

I do not see a return to per-interface/per-property controller calls which was causing performance issues in this modules a long time back so just double checking. There are, however, two areas worth considering for larger fabrics.

#### 1. Precompute binding indexes

Several GIE helpers repeatedly scan the complete binding registry or rebuild the same sets. The registry currently contains approximately 228 bindings, and these operations can occur repeatedly for every interface and property.

Consider constructing immutable indexes once when the module is loaded, for example:

- `(parent_template, profile_key) -> binding`
- `parent_template -> bindings`
- Precomputed sets for all registered, guarded, and no-log keys

This would make individual binding resolution an O(1) dictionary lookup and limit parent-level iteration to that parent’s bindings. It would not change behavior, but it would prevent the cost from increasing unnecessarily as both the interface count and registry grow.

#### 2. Restrict general carry-forward to `merged`

Generic carry-forward currently applies to `merged`, `replaced`, and `overridden`. Carry-forward means copying registered properties from `have` into the outgoing payload when they were omitted from `want`.

For example, assume the existing interface contains:

```yaml
description: old-description
mtu: 9216
speed: 100G
```

and the user supplies:

```yaml
description: new-description
```

For `state: merged`, preserving `mtu` and `speed` is expected because the user requested a partial update.

For `state: replaced` or `state: overridden`, silently carrying `mtu` and `speed` forward is generally inconsistent with the state semantics. Omitted managed properties should normally be reset or returned to their defaults. Carrying every registered property forward makes these states behave more like `merged`.

There may be a small number of controller-required or fabric-owned fields that must be retained in every state. Those should be explicitly identified and preserved through a narrow allowlist rather than applying general carry-forward to every registered binding.

This also has a scale implication. The routed-interface parent currently has up to 56 registered bindings. For 1,000 interfaces, carry-forward can inspect and potentially serialize tens of thousands of additional `nvPairs`, increasing:

- Local comparison and payload-construction time
- Memory and JSON serialization cost
- Request-body size
- NDFC parsing and template-processing work
- Debug logging cost when complete expanded payloads are formatted

A local engine-only synthetic check showed the relevant routed-profile processing at approximately 0.3 seconds per phase for 1,000 interfaces and 1.5–2 seconds per phase for 5,000 interfaces. These measurements are directional rather than end-to-end—controller latency will normally dominate—but they demonstrate that the current lookup pattern scales linearly and will become more noticeable as the registry grows.

I recommend:

1. Precomputing the registry indexes and key sets.
2. Applying general carry-forward only for `state: merged`.
3. Explicitly allowlisting any fields that genuinely must survive `replaced` or `overridden`.
4. Adding a synthetic 1,000–5,000-interface test that verifies bounded controller calls, payload size, and processing time.

One additional tradeoff is that failed or malformed controller reads can now be retried up to three times. That does not affect the normal success path, but it can increase latency and request volume when multiple switches are unreachable.

dacasti2mx:
Thanks, @mikewiebe  Agreed. We’ll address this in separate commits: 
-registry indexing, 
- bounded read recovery
- and withdrawal semantics, preserving fabric-owned fields. 

Each step will have measurable acceptance criteria and regression tests.

